PR #5, opened 2015-10-24
Specify rust target

phil-opp:


phil-opp:
Adding it to `setup_rust` and `printing_to_screen` tags.

PR #15, opened 2015-10-27
Add the clear_screen function from the blog post

phil-opp:


phil-opp:
Updated the `printing_to_screen` tag

Issue #17, opened 2015-10-29
1GB pages are not supported in QEMU<2.1 and Intel CPUs before 2010


phil-opp:
Support for 1GB pages was [introduced in QEMU 2.1] (http://wiki.qemu.org/ChangeLog/2.1) in 2014. Intel CPUs support it since [Westmere](https://en.wikipedia.org/wiki/Westmere_(microarchitecture)#Technology) (2010).

Either add a test through [CPUID 0x80000001, EDX bit 26](https://en.wikipedia.org/wiki/CPUID#EAX.3D80000001h:_Extended_Processor_Info_and_Feature_Bits) or use 2MB pages instead.

PR #18, opened 2015-10-29
Use 2MiB pages instead of one 1GiB page

phil-opp:
1 GiB pages are a relatively new feature, Intel introduced it in 2010 and QEMU in 2014. So we don't want to rely on it and make the whole OS incompatible to older systems. See issue #17.


phil-opp:
**TODO**: Cherry pick to tags: entering_longmode, setup_rust, printing_to_screen

PR #21, opened 2015-10-31
Use unsigned `mul` instead of signed `imul`

### LINKED ISSUE #20: Why use imul when computing table addresses?
https://github.com/phil-opp/blog_os/blob/entering_longmode/src/arch/x86_64/boot.asm#L57

This line uses imul, when unsigned semantics are required, in general.

phil-opp:
closes #20


phil-opp:
Cherry-picked to entering_longmode, setup_rust, and printing_to_screen

PR #43, opened 2015-11-15
Fix typo in "Entering long mode"

alevy:
I think there is a typo in the section describing how the P2 mapping is set up. The corresponding code sets _P2_'s third entry.


phil-opp:
Thanks!

PR #50, opened 2015-11-19
Don't set `write protect` bit when enabling paging

### LINKED ISSUE #47: Setting unneeded `write protect` bit when enabling paging
Ahmed Charles noted in [a comment](http://os.phil-opp.com/entering-longmode.html#comment-2367157772) that we set the `write protect` bit when enabling paging. It's bit 16 in the `CR0` register, [Wikipedia](https://en.wikipedia.org/wiki/Control_register#CR0) describes it like this:

> When set, the CPU can't write to read-only pages when privilege level is 0

The question is wether we should keep it (and add an explaining comment) or remove it (since it's not required for the switch to long mode).

I think it's reasonable to set it but maybe it's better to do it in a later post when it's required. When we remove it, it must not cause problems for people with the old version.

Any opinions are appreciated :).

phil-opp:
Don't set the `write protect` bit as we don't need it now. Resolves #47 .

### DIFF
--- a/posts/2015-08-25-entering-longmode.md
+++ b/posts/2015-08-25-entering-longmode.md
@@ -316,7 +316,6 @@ enable_paging:
     ; enable paging in the cr0 register
     mov eax, cr0
     or eax, 1 << 31
-    or eax, 1 << 16
     mov cr0, eax
 
     ret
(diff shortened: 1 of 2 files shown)


phil-opp:
Cherry-picked to `entering_longmode`, `setup_rust`, `printing_to_screen`, and `allocating_frames`

PR #67, opened 2015-12-13
Fix grammar: `set up` instead of `setup`

phil-opp:
`set up` instead of `setup`

Fixes #66

PR #69, opened 2015-12-13
Make comment in code example shorter

phil-opp:


phil-opp:
Accidentally included in #67

PR #85, opened 2015-12-23
Add unsafe to Unique::new

alicemaz:
previously compile would fail with E0133


alicemaz:
this brings the post in line with https://github.com/phil-opp/blog_os/blob/c19d176aa016383b011c7982c12bce6d55b8ff10/src/vga_buffer.rs#L25

phil-opp:
Thanks a lot!

PR #103, opened 2016-01-02
Use `-cdrom` instead of `-drive` as it's shorter

### LINKED ISSUE #101: Use `-cdrom` for QEMU
Replace

```
qemu-system-x86_64 -drive format=raw,file=os.iso
```

with the simpler

```
qemu-system-x86_64 -cdrom os.iso
```

See https://github.com/intermezzOS/book/commit/ac18e0d138ce6df465b8ab19061b4a9444a1ff58

phil-opp:
Right now, it does not matter whether we boot from a drive or a cdrom. Therefore we use the shorter `-cdrom`. Note that `-hda` would generate a warning in newer QEMU versions (see #56).

Fixes #101

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,10 +28,10 @@ clean:
 	@rm -rf build
 
 run: $(iso)
-	@qemu-system-x86_64 -drive format=raw,file=$(iso) -s
+	@qemu-system-x86_64 -cdrom $(iso) -s
 
 debug: $(iso)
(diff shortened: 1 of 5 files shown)


phil-opp:
Cherry picked to 

- [ ] multiboot_bootstrap
- [ ] entering_longmode
- [ ] set_up_rust
- [ ] printing_to_screen
- [ ] allocating_frames
- [ ] page_tables
- [ ] remap_the_kernel

PR #106, opened 2016-01-05
Add building instructions and links to posts and source code in README

phil-opp:


phil-opp:
Build fails because of rust-lang-nursery/rand#91 and rust-lang/rust#30713

PR #108, opened 2016-01-09
Fixed typos in post Allocating Frames

jonfk:
Small typos I found while reading and working through your posts. Thanks for the great work. I hope to see more coming, I have been learning a lot from these.


phil-opp:
Thanks a lot! I'm short of time at the moment, so probably there won't be new posts in the next month. But I will absolutely continue this :)

PR #120, opened 2016-01-19
Fix type name: `EntryFlags` instead of `TableEntryFlags`

phil-opp:
[Reported by FreeFull](http://os.phil-opp.com/modifying-page-tables.html#comment-2466158032)

### DIFF
--- a/posts/2015-12-09-modifying-page-tables.md
+++ b/posts/2015-12-09-modifying-page-tables.md
@@ -118,7 +118,7 @@ bitflags! {
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


FreeFull:
Hm, doesn't it actually make more sense for it to be `Entry::flags` rather than `EntryFlags::flags`? Having the method on `EntryFlags` would either be a no-op or not work.

phil-opp:
You're completely right, it should be `Entry::flags`

phil-opp:
Would you mind sending a quick pull request? I'm on mobile right now and can't fix it.

FreeFull:
Will do.

Issue #98, opened 2015-12-30
Extend some explanations


phil-opp:
Some issues that were brought up by rylev on IRC:
- [x] "CPUID check" too fast
- [x] more assembly explanations, especially `cmp` and `jmp`

_Update_: below @mtnygard mentioned some issues with  _Printing to Screen_:
- [ ] too quick
- [ ] unclear code fragments (which file? function or `impl` method?)
- [ ] better document the `lib.rs` and `Cargo.toml` changes required for `Unique` and `spin`
- [ ] type conflict: byte strings vs definition of `write_str`

mtnygard:
The part titled "Printing to Screen" also goes very quickly. The code fragments could use some more context: which file does it go in, is it top-level or inside a type? (E.g., the text doesn't state that write_str should be inside the Writer type definition.)

Using Unique and sping also require changes to the crate attributes and Cargo.toml. The text could use some clarification there as well.

mtnygard:
Section "Printing Strings" talks about adding ```b``` to make something a byte string, but then there is a type conflict with the definition of write_str.

I suspect this part assumes more Rust knowledge than I have here on day 3 of my journey. :-)

phil-opp:
Thanks a lot! I'll try to clarify these things.

henrikhodne:
I was going through the CPUID check myself to try to understand what was going on, and here's the version I ended up with, with some different comments:

``` ASM
check_cpuid:
	; Check if CPUID is supported by attempting to flip the ID bit (bit 21) in
	; the FLAGS register. If we can flip it, CPUID is available.

	; Copy FLAGS in to EAX via stack
	pushfd
	pop eax

	; Copy to ECX as well for comparing later on
	mov ecx, eax

	; Flip the ID bit
	xor eax, 1 << 21

	; Copy EAX to FLAGS via the stack
	push eax
	popfd

	; Copy FLAGS back to EAX (with the flipped bit if CPUID is supported)
	pushfd
	pop eax

	; Restore FLAGS from the old version stored in ECX (i.e. flipping the ID bit
	; back if it was ever flipped).
	push ecx
	popfd

	; Compare EAX and ECX. If they are equal then that means the bit wasn't
	; flipped, and CPUID isn't supported.
	xor eax, ecx
	jz .no_cpuid
	ret
.no_cpuid:
	mov al, "1"
	jmp error
```

mtnygard:
[...]

PR #127, opened 2016-02-01
Add/extend explanations in Entering Longmode

phil-opp:
Fixes some problems in #98


rylev:
Looks really nice. The only thing I think that could be helpful is just to mention that the CPUID checks are a bit "arbitrary", meaning that this is just how the X86 standard is set up. If it all seems a bit "hacky", it's because things at this level often do.

phil-opp:
Good suggestion, thanks! :)

PR #130, opened 2016-02-04
Improve introduction for frame allocation and explain what a frame is

### LINKED ISSUE #116: What is a frame?
Your blog series is normally easy to follow, but I got a bit lost when I started reading "Frame Allocation" because you leapt straight in without saying what a frame is. Could you dedicate a paragraph or two to this so the reader knows where they're standing?

phil-opp:
cc @Ketsuban (do you think it's better now?)

Closes #116


Ketsuban:
Yes, that's much clearer. Thanks very much.

PR #146, opened 2016-04-23
Add note that `stack_bottom` needs to be above `stack_top` in the source

### LINKED ISSUE #133: Add note that `stack_bottom` needs to be above `stack_top` in the source
Add a note in _Entering Longmode_ that `stack_bottom` needs to be _above_ `stack_top` in the source code because swapping these symbols causes [ugly bugs](http://os.phil-opp.com/entering-longmode.html#comment-2521183025).

phil-opp:
Fixes #133 

I'm not quite satisfied with the phrasing. I'm open to any suggestions…

cc @anula

### DIFF
--- a/posts/2015-08-25-entering-longmode.md
+++ b/posts/2015-08-25-entering-longmode.md
@@ -54,6 +54,8 @@ stack_bottom:
     resb 64
 stack_top:
 ```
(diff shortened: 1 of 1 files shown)


Evrey:
Wouldn't it be better to refactor those names to `stack_begin` and `stack_end`? This cannot be confused for a direction where one might assume that `top` has a higher address than `bottom`.

phil-opp:
The problem is that the stack grows downwards. So the end of the stack comes before its begin, which is counterintuitive too..

phil-opp:
Closing for the reasons outlined in #133

PR #147, opened 2016-04-23
Remove wrong claims about the linux allocator

### LINKED ISSUE #144: Correct information about Linux's internal allocator
Linux uses the buddy allocator for _frame_ allocations, not for virtual memory. Thanks to @DSMan195276 for [pointing it out](https://www.reddit.com/r/rust/comments/4ef3xv/writing_an_os_in_rust_kernel_heap/d1zqnla)!

We should fix the last paragraph of the introduction and also remove the wrong “kernels normally don't do heavy concurrent allocations” claim.

phil-opp:
Fixes #144 

cc @DSMan195276 (thanks again for your detailed comments on /r/rust)


DSMan195276:
I'd say this is perfectly fine. No need to go into detail on the Linux allocator. As far as allocators go it's not really any different then other allocators, and really the interesting parts related to it being a kernel allocator are much more complex and nuanced, definitely not something to bring up in this post.

phil-opp:
Thanks for reviewing, @DSMan195276!

PR #154, opened 2016-05-03
Add missing `use` statements in printing-to-screen

worr:
Although minor, I've added some missing `use` statements to the code examples in the VGA buffer entry


phil-opp:
Thanks a lot!

PR #162, opened 2016-05-21
remove zero-extension warning

benaryorg:
That nasm warning annoyed me. A lot.

### DIFF
--- a/src/arch/x86_64/boot.asm
+++ b/src/arch/x86_64/boot.asm
@@ -207,4 +207,7 @@ gdt64:
     dq (1<<44) | (1<<47) | (1<<41) ; data segment
 .pointer:
     dw $ - gdt64 - 1
-    dq gdt64
+    ; we actually need `dq` but it would throw a warning because of zero extension
+    dd gdt64
+    ; so we extend ourselves
+    dd 0


phil-opp:
I don't see any assembly warning with the current code… What version of nasm are you using?

benaryorg:
I'm using NASM version 2.12.01.

benaryorg:
I wondered why you had this particular piece of code in there anyway.

phil-opp:
Oh, I'm using 2.11.08. It seems like the warning was [added in 2.12](http://www.nasm.us/doc/nasmdocc.html). So let's get rid of this warning :).

> I wondered why you had this particular piece of code in there anyway.

What do you mean?

phil-opp:
I think we don't even need `dq` here. We are still in 32-bit mode, so the pointer structure should only need a 32-bit address…

benaryorg:
> What do you mean?

I wondered why you left a bit of code in there that threw a warning.

> I think we don't even need `dq` here. We are still in 32-bit mode, so the pointer structure should only need a 32-bit address…

I didn't really read the part about the GDT's as I was to reread everything when the OS had the basic Rust stuff running so I could (as with paging) setup everything once again.
I just added a second bit of space in case you actually used `dq` because it was a given structure.

benaryorg:
If you say that the field isn't actually used I could just update the PR.

benaryorg:
Oh wait, no I'm getting *that*:
```text
build/arch/x86_64/boot.o: In function `gdt64.pointer':
src/arch/x86_64/boot.asm:(.rodata+0x1a): relocation truncated to fit: R_X86_64_16 against `.rodata'
Makefile:49: recipe for target 'build/kernel-x86_64.bin' failed
make: *** [build/kernel-x86_64.bin] Error 1
```
if I remove the second `dw`.

phil-opp:
The [osdev wiki](http://wiki.osdev.org/Global_Descriptor_Table) and the [AMD64 manual](http://developer.amd.com/wordpress/media/2012/10/24593_APM_v21.pdf) agree: 2 bytes size and 4 bytes offset. I just tried it with `dd` instead of `dq`, and it works fine for me.

benaryorg:
I have just cloned a fresh copy of your repo and compiled (master branch) and got the above error while linking when replacing `dq` with just one `dd`.

That's strange.

I'm using `GNU ld (Gentoo 2.25.1 p1.1) 2.25.1` for linking.

phil-opp:
Weird. So nasm complains when you're using `dq` because it is only a 32-bit address and ld complains when you're using `dd` because the 32-bit address does not fit into the doubleword? And it works when you add another `dd 0` after it? Makes no sense…

benaryorg:
You are using that exact address to store the pointer to jump to in 64-bit mode so it has to be 64-bit long, but nasm complains about the `dq gdt` statement because `gdt` is a 32-bit pointer and you are trying to save it into a 64-bit register and it warns that it is being padded with 0s, but if you use a 32-bit element to store the address and add the 0s yourself then everything's fine.

phil-opp:
I've just tried it with nasm 2.12.01 and I can't reproduce the warning… I get the following for both `dd gdt64` and `dq gdt64`:

```
> nasm -version
NASM version 2.12.01 compiled on May 24 2016
> nasm -felf64 src/arch/x86_64/boot.asm
> ld src/arch/x86_64/boot.o
ld: warning: cannot find entry symbol _start; defaulting to 00000000004000b0
src/arch/x86_64/boot.o: In function `start':
src/arch/x86_64/boot.asm:(.text+0x37): undefined reference to `long_mode_start'
```

----

I tried some other (wrong) configurations:

With `dw gdt64`:

```
> nasm -felf64 src/arch/x86_64/boot.asm
> ld src/arch/x86_64/boot.o
ld: warning: cannot find entry symbol _start; defaulting to 00000000004000b0
src/arch/x86_64/boot.o: In function `start':
src/arch/x86_64/boot.asm:(.text+0x37): undefined reference to `long_mode_start'
src/arch/x86_64/boot.o: In function `gdt64.pointer':
src/arch/x86_64/boot.asm:(.rodata+0x1a): relocation truncated to fit: R_X86_64_16 against `.rodata'
```
Which is identical to your linker error.

--

With `-felf` instead of `-felf64` and `dq gdt64`:

```
> nasm -felf src/arch/x86_64/boot.asm 
src/arch/x86_64/boot.asm:210: warning: 64-bit unsigned relocation zero-extended from 32 bits
```
Which I think is identical to the warning you described. But with this configuration, I get the following linker error:

```
ld: i386 architecture of input file `src/arch/x86_64/boot.o' is incompatible with i386:x86-64 output
```

----

Are you sure that you've set these parameters correctly? And could you post your Makefile and your `boot.asm`?

benaryorg:
I've actually just used the provided Makefile, which works fine, except for the nasm warning.

I'm also running a current x86_64 Gentoo GNU/Linux.

phil-opp:
> I've actually just used the provided Makefile

Yeah, I've tried that too. But no nasm warnings or linker errors for me, neither for `dq gdt64`, nor for `dd gdt64` (with nasm 2.12.01)…

benaryorg:
That's confusing.

I'll just close this PR and just in case you happen to get those errors, just reopen/comment and I will be there to `git rebase` so you can merge it then.

phil-opp:
I would be happy to merge just the `dq gdt64` -> `dd gdt64` change, since it's more correct. However, it seems to make the situation even worse for you (linker error instead of warning), and I don't understand why…

> just in case you happen to get those errors, just reopen/comment and I will be there to git rebase so you can merge it then.

Thanks a lot! And thank you for reporting this issue.

--

@steveklabnik I think you're using the [same GDT code](https://github.com/intermezzOS/kernel/blob/master/src/src/arch/x86_64/boot.asm#L88) in intermezzOS. Do you experience this nasm warning or linker error too (with nasm 2.12)?

steveklabnik:
I am! But I ended up downgrading to nasm 2.11, because it choked on some macro code that I had. I'm in the process of removing those macros, so should be able to try again with 2.12 soon.

PR #176, opened 2016-06-07
Update section header to reflect changes in the text

steveklabnik:
Someone in `#rust-osdev` was following along, and while you no longer use `-Zno-landing-pads`, the title still talked about it. I don't know if you wanted to keep it, or use something else, but here's a suggestion :smile:


phil-opp:
Thanks a lot!

PR #178, opened 2016-06-12
Remove unneeded gcc_except_table section

phil-opp:
No longer needed because we use `panic=abort` now.


phil-opp:
Cherry picked to remap_the_kernel, kernel_heap, and catching_exceptions.

PR #187, opened 2016-06-25
Catch divide error instead of page fault

### LINKED ISSUE #184: Catch Divide-By-Zero faults instead of Page Faults in “Catching Exceptions” Post
Currently, we try to catch page faults using a simple `extern "C" fn() -> !` handler function. The problem is that the CPU pushes an error code on page faults, which leads to a wrong stack pointer alignment. Thus the final code causes a general protection fault on real hardware (or in QEMU with `-enable-kvm`) (reported by @nxnfufunezn in #179).

Possible solutions:

1. Insert a random `push 0` in the handler, without a detailed explanation.
2. Add a `Testing on real Hardware` section at the end of [Catching Exceptions](http://os.phil-opp.com/catching-exceptions.html), where we discover the bug. Debug the GP fault in the upcoming post.
3. Avoid the GP fault by catching Divide-By-Zero error (DE) instead of page faults. The DE doesn't push an error code, so the stack is correctly aligned.

I don't like option 1, since it's only a hacky workaround. Option 2 is fine and emphasizes the differences between QEMU and real hardware (QEMU doesn't check alignment on `movaps` instructions). However, newer QEMU version might do the alignment check and then our code doesn't work anymore.

So I think that option 3 is the best solution.

phil-opp:
The divide error pushes no error code. Thus we avoid stack misalignment.

Fixes #184

### DIFF
--- a/blog/post/2016-05-28-catching-exceptions.md
+++ b/blog/post/2016-05-28-catching-exceptions.md
@@ -1,9 +1,10 @@
 +++
 title = "Catching Exceptions"
 date = "2016-05-28"
+updated = "2016-06-25"
 +++
 
-In this post, we start exploring exceptions. We set up an interrupt descriptor table and add handler functions. At the end of this post, our kernel will be able to catch page faults.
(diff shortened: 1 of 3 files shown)


phil-opp:
Cherry-picked to `catching_exceptions`

PR #190, opened 2016-07-14
Add aligned .got and .got.plt sections to linker script

phil-opp:
These sections suddenly appeared on a newer compiler…


phil-opp:
Cherry-picked to remap_the_kernel, kernel_heap, catching_exceptions

PR #191, opened 2016-07-17
Display a Table of Contents on large displays

phil-opp:
# Option 1

The ToC is displayed through javascript on the left side when the display is large enough:

--

screenshot at 2016-07-17 16 46 46

---

The ToC position is fixed so that it doesn't scroll with the rest of the page and remains always at the same place:

--

screenshot at 2016-07-17 16 48 35

---

On smaller displays, the ToC is hidden via `display: none`.

# Option 2
Put it on the right side and use absolute instead of fixed positioning:

--

screenshot at 2016-07-17 19 22 14

---

screenshot at 2016-07-17 19 22 35

# Option 3
Put it before the text, so that it is displayed on all devices.

# Unresolved Questions
- Other options?
- Is a ToC too much visual clutter?
- Design improvements?
- Should we also display it on smaller displays?
- Fade the ToC out until the mouse hovers it again?

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -54,6 +54,11 @@
           prefix: 'toc'
       });
 
+      var heading = document.createElement("H2");
+      var heading_text = document.createTextNode("Table of Contents");
+      heading.appendChild(heading_text);
+
+      container.appendChild(heading);
       container.appendChild(toc);
     }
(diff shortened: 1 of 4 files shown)


bjorn3:
Please make it fixed.

PR #193, opened 2016-07-27
Fix hex constant

ahstro:
`2f4f` and `2f4b` switches places in one places where it's mentioned, I assume this was a typo since all the others looked this way :)
Edit: And, now that I've managed to build the thing (thanks by the way), because `0x2f4b2f4f` is `OK` while the reversed is `KO`


phil-opp:
Good catch! Thanks a lot!

(The `2f` is the color (white on green) and `4b` and `4f` are the ASCII codes for `O` and `K`. So it prints `KO` if we swap `4b` and `4f`.)

ahstro:
Yeah, I figured that out at 2<sup>am</sup> after three hours of getting this to compile. Thank you for this!

phil-opp:
You're welcome!

> after three hours of getting this to compile

That's a long time… What was the problem? Do you have any improvement suggestions for the description?

ahstro:
I only had a Macbook at hand, so I spent a lot of time trying different solutions. Ended up writing the code in a synced folder in a Vagrant box so I could compile the iso there but run it through VirtualBox on in OS X because qemu refused to work :sweat_smile: So nothing wrong with the tutorial, I just didn't have the appropriate tools at hand.

PR #195, opened 2016-08-03
Add rsi to the clobber list

Lewn:
In the second macro (handler_with_error_code), we pop the error code into rsi, this means we are changing the value of rsi in the assembly block, so we should add it to the clobbers.

I am quite new to this, so please close this request if I am completely wrong.

(Side question: An other thing I came across while trying to figure out if this was correct, is that every documentation says you should add curly brackets around register names, can someone tell me why this is not done here?)

### DIFF
--- a/blog/post/2016-08-03-better-exception-messages.md
+++ b/blog/post/2016-08-03-better-exception-messages.md
@@ -509,7 +509,7 @@ macro_rules! handler_with_error_code {
                       call $0"
                       :: "i"($name as extern "C" fn(
                           *const ExceptionStackFrame, u64) -> !)
-                      : "rdi" : "intel");
+                      : "rdi,rsi" : "intel");
                 ::core::intrinsics::unreachable();
             }
         }


phil-opp:
Good catch, thanks a lot!

phil-opp:
The curly braces around clobbers are wrong. It was discovered and made an error in https://github.com/rust-lang/rust/pull/34682

It is fixed in the nightly documentation at https://doc.rust-lang.org/nightly/book/inline-assembly.html but still there in in the stable and beta docs at https://doc.rust-lang.org/book/inline-assembly.html. Maybe this qualifies for a backport, since this is clearly wrong…

Is there any other other wrong documentation?

phil-opp:
It seems like doc fixes are not backported: https://github.com/rust-lang/rust/pull/33009#issuecomment-211481423.

So I will link to the nightly docs instead.

phil-opp:
Oh, it turns out that the syntax was not correct. I've opened #196 as follow-up. Thanks anyway!

PR #196, opened 2016-08-03
Fix clobber syntax and add missing rsi clobber in code

phil-opp:
Follow-up on #195


phil-opp:
Cherry-picked to `better_exception_messages`

PR #197, opened 2016-08-04
removed extra "use"

themasch:
I'm not a native speaker so I'm not 100% sure but I guess this "use" shouldn't be there.


phil-opp:
Thanks!

PR #204, opened 2016-08-09
Fix typo (therefor -> therefore)

Lewn:
I am not a native English speaker, so I am not 100% sure this is correct, but from the information I found on https://www.translegal.com/legal-english-lessons/therefore-vs-therefor , it looks like in this case, the word "therefore" is correct.


phil-opp:
Thanks a lot! I'm not a native speaker either, so I'm not 100% sure about this... 

The `remap-the-kernel` change looks good to me, but I think we should reword the sentence in `page-tables`. Something like "We implement it using the `FrameAllocator` from..." or "For the implementation we need the `FrameAllocator` from...". What do you think?

Lewn:
I think that "For the implementation we need the FrameAllocator from..." would read the easiest.

phil-opp:
Ok, thanks! Could you include it in the PR?

Lewn:
Done :)

phil-opp:
Thanks a lot!

PR #208, opened 2016-08-20
Fix various typos

KarboniteKream:
While going through the posts, I found some small typos:

1. In `boot.asm`, the `start` function was `_start` in one listing,
2. The `Set up Rust` post had an incorrect year of last update (2015 instead of 2016),
3. `unwrap` instead of `unmap` in `Modifying Page Tables`,
4. Changed `reexport` to `re-export` in some cases.


phil-opp:
Thanks so much!

PR #221, opened 2016-09-22
Use rustup for cross compiling and also explain xargo

### LINKED ISSUE #215: Suggestion: Cross complie with xargo
In the blog is a 'Cross Compiling: Libcore' post which describes compiling libcore to `x86_64-unknown-linux-gnu' for the os to compile. This is tedious, and the suggested "easy method" of using cargo requires changing _every_ crate that uses core. My suggestion is to use [xargo](https://github.com/japaric/xargo) to cross compile instead.

The main pros of using xargo in my opinion are as follows
+ It's easy: All I had to do to create cross compiling in my kernel was change `target` in the makefile to 'x86_64-unknown-none-gnu', replace all instances of cargo to xargo, and leave a [x86_64-unknown-none-gnu.json](https://gist.github.com/4e554c4c/1c3839545631308b49faaa2e5b088774) file in the project root.
+ It is scaleable: The llvm target file allows for easy later changes, such as the disabling of sse and hard floats that I did in mine.
+ Installing is not difficult: One can compile xargo on their machine with `cargo install xargo` if they are running a nightly compiler

However, there may be several cons as well
+ Xargo means another requirement for someone to run blog_os. Every person who would compile blog_os from the point this was imlemented onward would have to have to have xargo on their computer.
 + A solution to this would be to put it in the additional resources section, however this would be incredibly difficult with something as integral to the project as a build tool. Every post would have to consider both the xargo and cargo methods of compilation which defeats the purpose
+ It could create a split in the userbase: There may be people who have been folowing the blog for some time and use the traditional method of compilation. There may be an exception post which assumes the kernel compiled with "disable-redzone" which would not work on user's existing kernels.
+ It would create another reliability on unstable software, something else that can break, and something else to support. 

I can create a pull request to implement this, however I think that there should be a conversation for it first, especially because of the different levels that xargo can be used in the project.

phil-opp:
Fixes #215 

cc @4e554c4c

### DIFF
--- a/blog/additional-resource/cross-compile-libcore.md
+++ b/blog/additional-resource/cross-compile-libcore.md
@@ -2,46 +2,32 @@
 title = "Cross Compiling: libcore"
 +++
 
-So you're getting an ``error: can't find crate for `core` [E0463]`` when using `--target x86_64-unknown-linux-gnu`. That means that you're not running Linux or not using using a x86_64 processor.
+If you get an `error: can't find crate for 'core'`, you're probably compiling for a different target (e.g. you're passing the `target` option to `cargo build`). Now the compiler complains that it can't find the `core` library. This document gives a quick overview how to fix this problem. For more details, see the [rust-cross] project.
 
-**If you have an x86_64 processor and want a quick fix**, try it with `x86_64-pc-windows-gnu` or `x86_64-apple-darwin` (or simply omit the explicit `--target`).
-
-The idiomatic alternative and the only option for non x86_64 CPUs is described below. Note that you need to [cross compile binutils], too.
-[cross compile binutils]: /cross-compile-binutils.html
+[rust-cross]: https://github.com/japaric/rust-cross
 
 ## Libcore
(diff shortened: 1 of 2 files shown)


blog/additional-resource/cross-compile-libcore.md
```diff
@@ -2,46 +2,32 @@
...
 
-## Build it using cargo
-The alternative is to use cargo to build libcore. But this variant has one big disadvantage: You have to modify each crate you depend on because it needs to use the same libcore. So you can't just add a crates.io dependency anymore, you need to fork and modify it first.
+## rustup
+Thanks to [rustup], cross-compilation for [official target triples] is pretty easy today: Just execute `rustup target add x86_64-unknown-linux-gnu`.
 
-If you want to build libcore anyway, you need its source code. You can either clone the [rust repository] \(makes updates easy) or manually [download the Rust source][Rust downloads] \(faster and less memory).
-[rust repository]: https://github.com/rust-lang/rust
+[rustup]: https://rustup.rs
+[official target triples]: [target triple]: https://github.com/japaric/rust-cross#the-target-triple
```
bjorn3: Because of the extra `[target triple]:` here it renders as plain text.

phil-opp: Thanks! Fixed in #228.

Issue #220, opened 2016-09-20
BumpAllocator allocate() function vunerable to overflow


4e554c4c:
This might be a complete non-issue because it would only happen in release mode, but I thought I should bring it up. Calling allocate(-n as usize, 0) would actually _free_ n bytes in the allocator and never run out of memory. A better option would be use a saturating add which would never free memory make the function simply return `None` when called with a negative number.

phil-opp:
Good catch! This should definitely be a saturating add. Thanks a lot for reporting!

Do you want to send a PR? Alternatively, I could fix it myself if you don't have the time…

4e554c4c:
Sure! If you are fine waiting a day, I will fix this.

phil-opp:
> If you are fine waiting a day

Of course! Thanks a lot!

PR #223, opened 2016-09-22
Fix for #220: BumpAllocator overflow

4e554c4c:
I also fixed a spelling error and replaced a tab with a space
in the blog post where `allocate()` was created.

### DIFF
--- a/blog/post/08-kernel-heap.md
+++ b/blog/post/08-kernel-heap.md
@@ -86,8 +86,8 @@ impl BumpAllocator {
 
     /// Allocates a block of memory with the given size and alignment.
     fn allocate(&mut self, size: usize, align: usize) -> Option<*mut u8> {
-    	let alloc_start = align_up(self.next, align);
-        let alloc_end = alloc_start + size;
+        let alloc_start = align_up(self.next, align);
+        let alloc_end = alloc_start.saturating_add(size);
 
         if alloc_end <= self.heap_start + self.heap_size {
             self.next = alloc_end;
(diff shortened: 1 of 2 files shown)


blog/post/08-kernel-heap.md
```diff
@@ -99,7 +99,7 @@ impl BumpAllocator {
 }
 ```
 
-The `heap_start` and `heap_size` fields just contain the start address and size of our kernel heap. The `next` field contains the next free address and is increased after every allocation. To `allocate` a memory block we align the `next` address using the `align_up` function (decribed below). Then we add up the desired `size` and make sure that we don't exceed the end of the heap. If everything goes well, we update the `next` address and return a pointer to the start address of the allocation. Else, we return `None`.
+The `heap_start` and `heap_size` fields just contain the start address and size of our kernel heap. The `next` field contains the next free address and is increased after every allocation. To `allocate` a memory block we align the `next` address using the `align_up` function (described below). Then we add up the desired `size` and make sure that we don't exceed the end of the heap. We use a saturating add so that the calling code cannot overflow `heap_start`, which would cause undefined behaviour. If everything goes well, we update the `next` address and return a pointer to the start address of the allocation. Else, we return `None`.
```
phil-opp: >... so that the calling code cannot overflow `heap_start`

I understand what you mean, but this formulation might be a bit confusing. Maybe just "so that `alloc_start` cannot overflow"?

phil-opp:
Thanks a lot for the PR! What do you think about my comment?

blog/post/08-kernel-heap.md
4e554c4c: `alloc_end` or `alloc_start`? `alloc_start` is not changing, I'm fine with the change though.

4e554c4c:
Do you want me to also add an 'Updates:' section like you have on other posts that details the change?

blog/post/08-kernel-heap.md
phil-opp: Oh sorry.. Yes, I meant `alloc_end`.

phil-opp:
Thanks a lot!

(I don't think that we need an "updates" section for that, since it doesn't break any code.)

phil-opp:
Cherry-picked to kernel_heap, catching_exceptions, better_exception_messages, and returning_from_exceptions.

PR #239, opened 2016-10-06
Printing to Screen: Use volatile operations for accessing the VGA Buffer

### LINKED ISSUE #168: Use volatile writes for the VGA buffer
Thanks to @Amanieu for the [suggestion on /r/rust](https://www.reddit.com/r/rust/comments/4lfx0f/writing_an_os_in_rust_catching_cpu_exceptions/d3n9r9w).

phil-opp:
This PR wraps the `ScreenChar`s in a `Volatile` wrapper in the `Buffer` definition. The `Volatile` wrapper ensures that all accesses are volatile so that the compile can't optimize the writes away. This makes the code a bit more verbose since we can no longer copy whole rows at once.

This PR also updates and restructures the corresponding blog post:

- There is now a new, quite large subsection about volatiles (only a first draft, probably needs some fine-tuning).
- The “Printing to Screen” section was split into “Printing Characters” and “Printing Strings”. 
- We now derive Debug, Clone, and Copy for more types after fixing the first _“cannot move out of borrowed content”_ error. Otherwise we would get an error when defining the character array in the Buffer type. This change avoids a second error later in the post.
- Added some explanations for the code of the `new_line` and `clear_row` functions

Finally fixes #168

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ multiboot2 = "0.1.0"
 once = "0.3.2"
 rlibc = "0.1.4"
 spin = "0.3.4"
+volatile = "0.1.0"
 
 [dependencies.hole_list_allocator]
 path = "libs/hole_list_allocator"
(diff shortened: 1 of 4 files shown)


phil-opp:
Cherry picked to printing_to_screen allocating_frames page_tables remap_the_kernel kernel_heap catching_exceptions better_exception_messages returning_from_exceptions

PR #245, opened 2016-10-18
Typofix

Dentosal:
Fixed typo in the kernel remapping chapter.


4e554c4c:
I think that "therefor" is preferable to "therefore" in that situation, but I guess you could look at the sentence two ways.
1) Here is a problem, for that we use our solution _
2) Here is a problem, so we must solve it with _
I think that sentence in question is more like the first one.

4e554c4c:
It's @phil-opp's call though, you might want to leave it open for him.

phil-opp:
Thanks for the PR @Dentosal!

I think we should replace the uncommon “therefor” with e.g. “for that” in this case.

PR #247, opened 2016-10-23
Fixed broken link in 07-remap-the-kernel.md

Dentosal:
A link to the source code of the `tlb::flush_all` function was broken.


phil-opp:
Thanks a lot!

PR #249, opened 2016-10-30
Update `print` macro to avoid deadlock; remove hacky `print_error` function

phil-opp:
This PR changes the implementation of the `print` macro as follows:

Old:

```rust
macro_rules! print {
    ($($arg:tt)*) => ({
           use core::fmt::Write;
           $crate::vga_buffer::WRITER.lock().write_fmt(format_args!($($arg)*)).unwrap();
    });
}
```

New:

```rust
macro_rules! print {
     ($($arg:tt)*) => ({
            $crate::vga_buffer::print(format_args!($($arg)*));
     });
}

pub fn print(args: fmt::Arguments) {
    use core::fmt::Write;
    WRITER.lock().write_fmt(args).unwrap();
}
```

With the new implementation, the formatting arguments are evaluated (through `format_args`) _before_ the global `WRITER` is locked. This has some advantages:

- `println!("{}", {println!(""); 1))` no longer causes a deadlock
- When an exception occurs while evaluating an argument, the global `WRITER` is still unlocked. Thus, the exception handlers can safely print to screen using a normal `println`. So we don't really need a hacky `print_error` function anymore.

This means that we can completety remove the `print_error` function and multiple unsafe blocks (since we can just use the safe println now). We can also remove the _“Exceptions inside println”_ section from “Catching Exceptions”.

The drawback is that errors are no longer printed in red. We could fix this by re-adding a safe `print_error` function. However, I think that's it's not worth the additional complexity for now. Instead, I plan to create a new VGA driver (with proper color support) as soon as we have keyboard input.

--

**TODO**:

- [x] Update “Printing to Screen”: introduce a new section about deadlocks.
- [x] Update “Catching Exceptions”: remove the section about exceptions inside println. 
- [x] Update “Better Exception Messages”
- [x] Update “Returning from Exceptions”
- [x] Update all screenshots that show the old red error messages. (implemented in #250)


phil-opp:
- [x] cherry-picked d96df500a7d7c199dc6206638b49fdfd682996c8 to printing_to_screen allocating_frames page_tables remap_the_kernel kernel_heap catching_exceptions better_exception_messages returning_from_exceptions
- [x] cherry-picked 9b83b2853e7182294107612514953a840aa2b149 to catching_exceptions better_exception_messages returning_from_exceptions

PR #252, opened 2016-11-01
Exceptions: Take the ExceptionStackFrame per reference

phil-opp:
This PR changes the type of exception handler functions. Instead of a `*const ExceptionStackFrame`, they now take a `&ExceptionStackFrame` as argument. The reason is that they require the pointer to be valid (they used to blindly derefence the raw pointer). So the old implementation was unsound, since the functions used to violate memory safety when an invalid pointer is passed.

A consequence of this change is that we should no longer modify the exception stack frame (since we have an immutable reference to it). And transmuting a `&` to a `&mut` is undefined behavior. So the last section of _Returning from Exceptions_ (_“Page Faults as Breakpoints”_) becomes even more hacky. Therefore, this PR removes that section. Instead, we now test the iretq logic by entering an endless loop between the exception handler and the instruction accessing 0xdeadbeaf.

### DIFF
--- a/blog/post/10-better-exception-messages.md
+++ b/blog/post/10-better-exception-messages.md
@@ -55,12 +55,11 @@ Now we need a way to find the memory address of this stack frame. When we look a
 // in src/interrupts/mod.rs
 
 extern "C" fn divide_by_zero_handler() -> ! {
-    let stack_frame: *const ExceptionStackFrame;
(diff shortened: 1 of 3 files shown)


phil-opp:
Cherry picked to better_exception_messages and returning_from_exceptions

PR #256, opened 2016-11-26
Fix panic_fmt arguments: file is a &'static str

phil-opp:


phil-opp:
Cherry picked to allocating_frames page_tables remap_the_kernel kernel_heap catching_exceptions better_exception_messages returning_from_exceptions

PR #257, opened 2016-11-29
Test for rustfmt/xargo instead of using `|| true`

phil-opp:


phil-opp:
cherry-picked to `returning_from_exceptions`

PR #246, opened 2016-10-22
map the P4 table recursively before setting paging

matanelevi:
We have to map the P4 table recursively before setting paging - can't do this wherever we'd like.

PR #259, opened 2016-11-30
Revert "map the P4 table recursively before setting paging"

phil-opp:
Gil left [this comment](http://os.phil-opp.com/modifying-page-tables.html#comment-3015227025) regarding #246 on the blog:

> You said that the P4 recursive loop must be set before paging is enabled.
But I wonder - the memory is currently identity mapped, so what difference is there in P4_table address in with/without paging?

I think this is correct. I also tested it and it still works when I do the recursive mapping after enabling paging.

Reverts phil-opp/blog_os#246


phil-opp:
Any objections, @matanelevi?

PR #272, opened 2016-12-29
Add a Xargo.toml for xargo 0.3.0 and update the corresponding post

### LINKED ISSUE #271: Add a Xargo.toml for xargo 0.3.0
The newest xargo version 0.3.0 only builds the core crate by default. In order to build `alloc` and `collections` as well we now need a file named `Xargo.toml` with the following contents:

```toml
[target.x86_64-blog_os.dependencies]
collections = {}
```

This seems to work, too (but I'm not sure if it's idiomatic):

```toml
[dependencies]
collections = {}
```

phil-opp:
Xargo 0.3.0 was just released: https://users.rust-lang.org/t/xargo-v0-3-0-released-build-your-own-std/8571

By default it compiles only `core` now. However, we also need `alloc` and `collections`. We can enable them by creating a file named `Xargo.toml` with the following content:

```toml
[target.x86_64-blog_os.dependencies]
collections = {}
```

Fixes #271

### DIFF
--- a/Xargo.toml
+++ b/Xargo.toml
@@ -1,0 +1,2 @@
+[target.x86_64-blog_os.dependencies]
+collections = {}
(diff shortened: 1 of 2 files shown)


phil-opp:
Cherry-picked to `returning_from_exceptions` tag.

PR #278, opened 2017-01-14
Remove the data segment from assembly 64-bit GDT

### LINKED ISSUE #273: Entering Long Mode: reloading ds, ss, and es unnecessary
According to the AMD/Intel manuals, the CPU complete ignores data segment registers in 64 bit mode.

### LINKED ISSUE #277: “Double Faults” post breaks returning from exceptions
“`int!(3)` produces a double fault while it returned correctly in the previous post”. ([reported by Blaž Šnuderl](http://os.phil-opp.com/double-faults.html#comment-3091388494))

It seems like the problem is our new GDT, which doesn't have a data segment descriptor. This alone is fine, but the `ss` register still holds a value which is saved to the stack when an exception occurs and checked by `iretq` when we return. If the saved value is not a valid data segment descriptor (or a null descriptor), a general protection fault occurs.

One possible solution is to load 0 into the `ss` register after entering long mode (null descriptors are explicitely allowed in `ss` in 64-bit mode). Another solution is to add a valid writable data segment to our new GDT.

I'm pretty sure that this is also the cause of the problem reported by @TengenJulian [on gitter](https://gitter.im/phil-opp/blog_os?at=586ed823aa6be0472f125c45):

>  I have been reusing the same GDT code for stack switching, but without the data segment, the iretq generates a GPF.

Related to #273

phil-opp:
The Intel and AMD manuals state that no data segment is needed in 64-bit mode. However, the `iretq` instruction expects a valid data segment selector or a null selector in the `ss` register.

This PR removes all ignored fields from the GDT structure overview in _Entering Long Mode_. It also removes the data segment from the assembly 64-bit GDT and no longer reloads the data segment registers before jumping to 64-bit mode. Instead, we reload them with null selectors after we're in 64-bit mode.

Fixes #273
Fixes #277

### DIFF
--- a/blog/post/02-entering-longmode.md
+++ b/blog/post/02-entering-longmode.md
@@ -383,45 +383,39 @@ start:
 To test it we execute `make run`. If the green OK is still printed, we have successfully enabled paging!
 
 ## The Global Descriptor Table
-After enabling Paging, the processor is in long mode. So we can use 64-bit instructions now, right? Wrong. The processor is still in some 32-bit compatibility submode. To actually execute 64-bit code, we need to set up a new Global Descriptor Table.
+After enabling Paging, the processor is in long mode. So we can use 64-bit instructions now, right? Wrong. The processor is still in a 32-bit compatibility submode. To actually execute 64-bit code, we need to set up a new Global Descriptor Table.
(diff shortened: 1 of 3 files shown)


phil-opp:
Cherry-picked e179dad and 8f8b46a to `entering_longmode set_up_rust printing_to_screen allocating_frames page_tables remap_the_kernel kernel_heap catching_exceptions better_exception_messages returning_from_exceptions double_faults`

Issue #304, opened 2017-03-30
Double Faults SegmentSelector::empty(), old x86 code


NullP01nt:
On the Double Faults blog post, you're using `SegmentSelector::empty()` in the `interrupts::init()` function.
However, on my version of `x86_64` (0.1.2), `SegmentSelector` has no member function called `empty()`, also according to the official docs. How would I initialize a SegmentSelector to be empty?

NullP01nt:
I've managed to solve this, looks like some of the code in the post was still dependent on the x86 crate. By replacing the call to `SegmentSelector::empty()` with `SegmentSelector::new(0, PrivilegeLevel::Ring0)`, you're effectively doing the same thing. Just don't forget to write `use x86_64::PrivilegeLevel;` somewhere appropriately in `src/interrupts/mod.rs`.

The code should be updated to complete the transition to the `x86_64` crate.

gil0mendes:
If you use just `SegmentSelector(0)` is actually the same thing, but a bit shorter. 🙂 And yes, you must put the `use x86_64::structures::gdt::SegmentSelector;` inside of the `init` function (you can also put it at the top of the page, but you won't use it anywhere else).

phil-opp:
Fixed in #305

PR #305, opened 2017-03-31
Fix empty selector

gil0mendes:
The “Double Fault“ post still use the old `SegmentSelector::empty()` from the x86 crate. This commit updates it to the new x86_64 crate.

This fix the issue described in #304


phil-opp:
Thanks a lot!

gil0mendes:
It's a pleasure to be able to help. 🙂

Issue #303, opened 2017-03-30
Handling Exceptions post outdated link and compilation errors


NullP01nt:
On the post about Handling Exceptions, where it links to the content on GitHub, it links to a branch/tag 'handling_exceptions' that no longer exists on the repository.

Further down, where it describes how to generate a breakpoint interrupt, my rust compiler tells me that `x86_64::instructions::interrupts::int3()` is not found, however, replacing it with x86_64's macro `unsafe { int!(3) };` works. This would require that the crate definition be marked with `#[macro_use]`.

I just upgraded to x86_64 version 0.1.2, which fixes the issue of finding the `int3()` function.

phil-opp:
Sorry for the late reply.

Thanks a lot for reporting! I was short on time when I published “Handling Exceptions” and decided to postpone the creation of the GitHub tag to the next day. But I still did not have the time to do it :disappointed: .

I'll try to fix both things over the easter weekend.

PR #311, opened 2017-04-13
Bring “Page Tables” post up to date again

phil-opp:
This PR:

- removes the testing of the `translate` method since the stack overflow no longer occurs (as we're using a 16KiB stack now)
- removes an unneeded unsafe block
- updates the x86_64 crate to version 0.1.2, which provides the `int3` function that we need in “Handling Exceptions” (this was reported in #303)


phil-opp:
Travis fails because of https://github.com/rust-lang-nursery/rustup.rs/issues/1062. It works locally, so I merge it.

PR #346, opened 2017-08-04
Use new Unique API

### LINKED ISSUE #345: Unique::new now returns Option
Ref: rust-lang/rust/pull/42959

`Unique::new` was changed to return an `Option<Self>`.
`Unique::new_unchecked` offers the same behavior as the old `new`.

I am up to [Printing Characters](https://os.phil-opp.com/printing-to-screen/#printing-characters). `print_something` does not compile using rustc 1.21.0-nightly (52a330969 2017-07-27) due to this change.

crawfxrd:
Change Unique to use `new_unchecked`. Fixes #345.


phil-opp:
Thanks!

PR #348, opened 2017-08-19
Update to use the new API for custom allocators.

too-r:
Had to change the linked_list_allocator dependecy to link directly to git, because the version on crates.io is still using the old `Unique` API.


4e554c4c:
Why are you creating `Cargo.lock`?

too-r:
@4e554c4c Cargo.lock is included in Git by default with this repository?

4e554c4c:
Then add it to the .gitignore. There isn't currently a `Cargo.lock` in the repository so you shouldn't add one without a good reason.

too-r:
Ok, fixed that.

too-r:
It seems I can't use Git apparently, `Cargo.lock` is properly removed now.

4e554c4c:
Thank you! I think this is good.

libs/hole_list_allocator/src/lib.rs
```diff
@@ -7,62 +7,51 @@
...
-        Heap::new(HEAP_START, HEAP_SIZE)
-    });
-}
+static HEAP: Mutex<Option<Heap>> = Mutex::new(None);
 
-#[no_mangle]
-pub extern fn __rust_allocate(size: usize, align: usize) -> *mut u8 {
-    HEAP.lock().allocate_first_fit(size, align).expect("out of memory")
+//Set up the heap
+pub unsafe fn init(offset: usize, size: usize) {
```
phil-opp: This init function isn't called right now, so the heap stays uninitialized.

too-r:
I think using the LockedHeap struct is a definite possiblity, however for now I have just changed it so the heap init function does get called in `src/memory/mod.rs` under the `memory::init` function.

phil-opp:
Ok, I decided to merge this for now. Maybe we switch to LockedHeap eventually, but for now it's good to have this project buildable again.

Thanks a lot @too-r!

src/lib.rs
```diff
@@ -53,7 +53,7 @@ pub extern "C" fn rust_main(multiboot_information_address: usize) {
 
     // initialize our IDT
     interrupts::init(&mut memory_controller);
-
+
```
bjorn3: trailing whitespace which was not previously here.

PR #361, opened 2017-10-20
A rewrite of the bump allocator to conform to the new API.

johanmon:
This version follows the new interface. I've tried to make the transition to the existing version of the hole_list_allocator easier so I've kept that pattern of having a init() function etc. 

If this version of the bump allocator is ok I could do a rewrite of the instructions in the blog.

### DIFF
--- a/libs/bump_allocator/src/lib.rs
+++ b/libs/bump_allocator/src/lib.rs
@@ -21,44 +21,86 @@ extern crate alloc;
 extern crate spin;
 
 
-static HEAP: Mutex<Heap> = Mutex::new(Heap::new(0, 0));
-
-//Set up the heap
-pub unsafe fn init(offset: usize, size: usize) {
-    *HEAP.lock() = Heap::new(offset, size);
-}
-
-#[derive(Debug)]
-struct Heap {
-    start: usize,
-    end: usize,
-    next: usize,
+struct LockedHeap {
+    heap: Mutex<Heap>,
 }
 
 
-impl Heap {
-    /// Initialisation of the heap to use the 
-    /// range [start, start + size).
-    const fn new(start: usize, size: usize) -> Heap {
-        Heap {
-            start: start,
-            end: start + size,
-            next: start,
+#[global_allocator]
+static GLOBAL_ALLOC: LockedHeap = LockedHeap::empty();
+
+
+pub unsafe fn init(start: usize, size: usize) {
+    GLOBAL_ALLOC.init(start, size);
+}
+
+
+impl LockedHeap {
+    // Creates an empty heap. All allocate calls will return 'AllocErr`.
+    pub const fn empty() -> LockedHeap {
+        LockedHeap {
+            heap : Mutex::new(Heap::empty())
         }
+    }
+
+    unsafe fn init(&self, start: usize, size: usize)  {
+        self.heap.lock().init(start, size);
     }
 }
 
-pub struct Allocator;
+// The interface used for all allocation of heap structures. 
 
-unsafe impl<'a> Alloc for &'a Allocator {
+unsafe impl<'a> Alloc for &'a LockedHeap {
 
     unsafe fn alloc(&mut self, layout: Layout) -> Result<*mut u8, AllocErr> {
-        let ref mut heap = HEAP.lock();
-        let alloc_start = align_up(heap.next, layout.align());
+        self.heap.lock().allocate(layout)
+    }
+
+    unsafe fn dealloc(&mut self, ptr: *mut u8, layout: Layout) {
+        self.heap.lock().dealloc(ptr, layout)
+    }
+}
+
+
+
+// A fixed size heap with a reference to the beginning of free space.
+
+pub struct Heap {
+    start: usize,
+    end: usize,
+    next:  usize,
+}
+
+impl Heap {
+    // Creates an empty heap. All allocate calls will return `AllocErr`.
+    pub const fn empty() -> Heap {
+        Heap {
+            start: 0,
+            end: 0,
+            next: 0,
+        }
+    }
+
+    // This is unsafe, the start address must be valid and the memory in
+    // the `[start, start + size)` range must not be used for anything
+    // else. This function is unsafe because it can cause undefined
+    // behavior if the given address or size are invalid.
+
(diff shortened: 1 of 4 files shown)


libs/bump_allocator/src/lib.rs
```diff
@@ -8,53 +8,75 @@
...
-    const fn new(heap_start: usize, heap_size: usize) -> BumpAllocator {
-        BumpAllocator {
-            heap_start: heap_start,
-            heap_size: heap_size,
-            next: heap_start,
+
+impl Heap {
+    /// Initialisation of the heap to use the 
+    /// range [start, start + size).
+    fn new(start: usize, size: usize) -> Heap {
```
phil-opp: We could mark this function as `const` to allow calling it in statics.

libs/bump_allocator/src/lib.rs
```diff
@@ -8,53 +8,75 @@
...
+extern crate alloc;
 extern crate spin;
 
+
 pub const HEAP_START: usize = 0o_000_001_000_000_0000;
 pub const HEAP_SIZE: usize = 100 * 1024; // 100 KiB
 
-static BUMP_ALLOCATOR: Mutex<BumpAllocator> = Mutex::new(
-    BumpAllocator::new(HEAP_START, HEAP_SIZE));
+static HEAP: Mutex<Option<Heap>> = Mutex::new(None);
```
phil-opp: Instead of wrapping `Heap` in an `Option`, we could initialize it with a zero sized heap (i.e. start==end). An allocation on an unitialized heap would then automatically fail instead of panicking.

libs/bump_allocator/src/lib.rs
johanmon: yes ( still learning rust :-)

libs/bump_allocator/src/lib.rs
johanmon: Make sens. 
What about moving the HEAP_START and HEAP_SIZE to the kernel lib.rs - makes more sense to have them there instead of in the allocator library?

phil-opp: I don't think that this is possible, since we can't use statics of a parent crate.

libs/bump_allocator/src/lib.rs
phil-opp: It's only needed if we remove the `Option` from the static `HEAP` (since we then need to construct a valid `Heap` at compile time).

johanmon:
The bump allocator as it is now needs to be initialized by the kernel. If this is so, the HEAP_START and HEAP_SIZE could move to the kernel lib.rs and this makes more sense. If the allocator should be initialized static then they should remain. I do like the static initialization but not that the location and size of the kernel heap is hard coded in a library. (might not know what I'm talking about :-)

johanmon:
One thing that struck me, the init(offset, size) function need not be unsafe does it?  I used hole_list_allocator as the target but there it is unsafe only since the linked_list_allocator uses an unsafe initialization, right?

phil-opp:
> The bump allocator as it is now needs to be initialized by the kernel. If this is so, the HEAP_START and HEAP_SIZE could move to the kernel lib.rs and this makes more sense.

Sounds good!

> One thing that struck me, the init(offset, size) function need not be unsafe does it?

The function implementation itself works fine without unsafe. The question is: Can we do something bad with the safe methods of a wrongly initialized allocator? If yes, we should mark `init`as unsafe, even if it doesn't do anything unsafe itself.

In this case, I'm not quite sure. We never write to any memory location (or do we?), so we just return invalid raw pointers on wrong initialization. So the unsafe steps should be just dereferencing the returned raw pointer or using the allocator as `global_allocator`, which are both outside of the interface. Or am I missing something?

bjorn3:
Calling init with wrong parameters is unsafe, as doing so will let the `(de)alloc` do unsafe things.

johanmon:
ok, HEAP_START and HEAP_SIZE in main lib.rs.  Leaving the init() unsafe.

phil-opp:
> Calling init with wrong parameters is unsafe, as doing so will let the (de)alloc do unsafe things.

Yes, but `alloc` and `dealloc` are unsafe itself, so this wouldn't be a problem. However, I agree that initializing the global allocator should be unsafe, since it is used implicitly by the compiler. So I'm in favor of leaving `init` unsafe, too.

phil-opp:
@johanmon What do you think about merging `Allocator`into `Heap`(see [my comment above] (https://github.com/phil-opp/blog_os/pull/361#pullrequestreview-71004730))?

johanmon:
Hmm, pushing my limits of rust here; I agree that having the Allocator as separate structure seams redundant but I can't turn it into using Heap only. 

If we have:
`#[global_allocator]
static GLOBAL_ALLOC: Heap = Heap::new(0,0);`

And implement Aloc for Heap:
`unsafe impl<'a> Alloc for &'a Heap {

    unsafe fn alloc(&mut self, layout: Layout) -> Result<*mut u8, AllocErr> {`

Then  the alloc function would modify self without any protection from the Mutex.lock. If we keep the static HEAP structure that is protected by a Mutex then elements of self are not used (?).  One could of course have a protected `next` value that is separate from the Heap structure but that would still be two structures.

The reason I wrote it as I did is that I wanted it to be identical to hole_list_allocator to make the transition easy.

phil-opp:
Oh sorry, I completely forgot about the Mutex. I would recommend to make `Allocator` a wrapper around heap: `pub struct LockedHeap(Mutex<Heap>)`. Then we only need one static instead of two.

johanmon:
ah,  ok I'll give it a try

johanmon:
ok,  how is this?

phil-opp:
Looks good, thanks for the update!

Just some stylistic fixes, then we can merge:

- In Rust, there is a special comment type for documenting items that is automatically used by rustdoc. A documentation comment is indicated by three slashes (`///`) instead of two on each line. Could you update the documentation comments to use this style?

- Could you remove the blank lines between the documentation comments and the start of the function?

johanmon:
Will fix.

phil-opp:
Thanks a lot for the pull request and the continuous updates, @johanmon!

PR #364, opened 2017-11-14
Fix ld error: Undefined symbol eh_personality.

too-r:
This can be fixed by telling the compiler not to mangle the function. Maybe something was changed so that you have to explicitly state to the compiler not the optimise the function away? I don't know. Point is, this builds for me now, whereas before cloning the repo fresh would throw up link errors.


phil-opp:
Thanks! I don't know why this is suddenly required either, but I try to investigate this. For now, no_mangle seems like the best solution.

PR #367, opened 2017-11-19
Update the “Kernel Heap” post to use new allocator API

### LINKED ISSUE #341: Global allocator API changed
See https://github.com/rust-lang/rfcs/pull/1974

**Update**: Build was fixed in #348. However, we still need to do some things:

- [x] adjust the blog post (done in #367)
- [x] backport the changes to all git tags
- [x] maybe get rid of the `hole_list_allocator` and use `LockedHeap` directly (also in #367)

phil-opp:
This PR rewrites the “Kernel Heap” post to the new global allocator API. It also moves the bump allocator inside the main crate (instead of being a separate crate in the `libs/` folder). Instead of using a spinlock for the global bump allocator, we use `AtomicUsize` to get a lock-free global allocator (of course, the updated post explains how it works).

I also updated the git tags for the various posts (e.g. `post_8` for “Kernel Heap”), so that they all build again.

Fixes #341


phil-opp:
cc @johanmon

Issue #360, opened 2017-10-19
Tracking Issue for Second Edition


phil-opp:
Create a second edition of the blog, which reorders the posts (exceptions before page tables) and uses an own bootloader. This should make the posts easier to follow. 

---

_Edit:_ This issue will serve as a **tracking issue**. The [“Second Edition” milestone](https://github.com/phil-opp/blog_os/milestone/1) contains all associated issues and pull requests.

## Planned high-level changes

- [x] Create a custom, dependency free bootloader
  - Written completely in inline assembly to be dependency free (see [this comment](https://github.com/phil-opp/blog_os/issues/360#issuecomment-339575121) for details)
  - First prototype: https://github.com/rust-osdev/bootloader
- [x] Work out a new order of posts
  - Remove the sudden jump in difficulty with the “Page Tables” post
  - My current plan: Setup & printing first, then exceptions and hardware interrupts (including keyboard and timer support, and maybe even multithreading), then memory management
- [x] Move more code into the `x86_64` crate
  - Remove low-level implementation details (e.g. how to align a memory address) from the blog text, but …
  - Document the code in the `x86_64` crate very well and ecourage readers to look at it

## Further changes

- [x] Link using LLD (#370)
  - LLD is about to be shipped with rustc: https://github.com/rust-lang/rust/pull/48125
- [x] “Kernel Heap”: include more allocator types
- ~~[ ] Put the kernel in the higher half of the address space (see https://github.com/phil-opp/blog_os/issues/360#issuecomment-350845368)~~ maybe sometimes in the future when we introduce linker scripts
- [x] Unit tests and other tests? (see https://github.com/robert-w-gries/rxinu/issues/32 for an overview of possibilities)
- [x] Be platform independent: The blog should work on Linux, Mac, and Windows
  - There are CI jobs that verify this.

## Other planned things

- [x] Use `rustfmt-nightly` so that the code always has standard format. Check formatting via travis.
- [x] Docker support (#369)
- [x] Make sure that every code snippet has a comment that specifies the file path (https://github.com/phil-opp/blog_os/issues/382).

## Maybe

- ~~[ ] Use a javascript emulator in blog posts (#368)~~ Postponed, since [there is no 64-bit js emulator yet](https://github.com/phil-opp/blog_os/issues/360#issuecomment-353765641)

---

(This list is far from complete and will be extended over time.)

robert-w-gries:
[...]

PR #372, opened 2017-12-04
[Fix #362] Fix sign extension bug in `next_table_address`

### LINKED ISSUE #362: `next_table_address` doesn't always yield a canonical address - sign extension bug
Firstly, thank you Phil for a great set of tutorials.

I've been trying to move the tutorial kernel into the higher-half, which has been interesting. In the `next_table_address` method of `Table`, we do something along the lines of:
```
let table_address = (self as *const _) as usize;
Some((table_address << 9) | (index << 12))
```
However, for some `table_address`s, this shift can mess up the sign extension required to make the address canonical, and so leads to General Protection Faults when trying to access the next page table.

I fixed this by manually re-extending the required bits:
```
let table_address = (self as *const _) as usize;
Some(((table_address << 9) | (index << 12)) | (0o177777_000_000_000_000_0000 * ((table_address>>47)&0b1)))
```

IsaacWoods:
Fix an issue where the left shift of the old table address would overwrite the sign extension, making the
address non-canonical and leading to #GPs. This calculates the correct sign extension for the new table
address.

### DIFF
--- a/src/memory/paging/table.rs
+++ b/src/memory/paging/table.rs
@@ -39,7 +39,8 @@ where
         let entry_flags = self[index].flags();
         if entry_flags.contains(PRESENT) && !entry_flags.contains(HUGE_PAGE) {
             let table_address = self as *const _ as usize;
-            Some((table_address << 9) | (index << 12))
+            let sign_extension = 0o177777_000_000_000_000_0000 * ((table_address >> 47) & 0b1);
+            Some(((table_address << 9) | (index << 12)) & ((1 << 48) - 1) | sign_extension)
         } else {
             None
         }
--- a/src/memory/paging/table.rs
+++ b/src/memory/paging/table.rs
@@ -31,6 +31,15 @@ where
     }
 }
 
+/*
+ * Addresses are expected to be canonical (bits 48-63 must be the same as bit 47), otherwise the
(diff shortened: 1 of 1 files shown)


IsaacWoods:
I chose to call it `make_address_canonical` instead, as it's specialised to address canonicalisation, rather than generic sign extension. Hope that is more readable.

Also, just something I was thinking about while implementing this, I think someone at some point mentioned the idea of replacing the aliases used in `VirtualAddress` and `PhysicalAddress` with the newtype pattern. I noticed that the addresses in the table methods use `usize`, which imo shouldn't be possible. If at some point you supported moving to a stricter use of the address types, I think that would increase safety and give a home to methods like `make_address_canonical`, and I'd be happy to do some of that legwork.

phil-opp:
Awesome, thanks! I like the function name.

Regarding the newtypes for VirtualAddress and PhysicalAddress: I definitely plan to do it for the upcoming second edition (#360). The plan is to move a lot of code to the x86_64 crate and define the newtypes there. The canonical address function would fit in there pretty well, too.

Changing the first edition would be a lot of work, since it touches a lot of code and we would need to adjust almost all posts. So I think it's a better idea to focus on the second edition instead.

PR #373, opened 2017-12-06
Docker: Add a Dockerfile and Makefile targets

Sh4d1:
Hey! 

I wrote a Dockerfile and an entrypoint (inspired by the Redox ones) to build the kernel via Docker. 

What do you think? 

Cheers :)


Makefile
```diff
@@ -33,6 +39,23 @@ run: $(iso)
...
 
+# docker_* targets 
+
+docker_build:
+	@docker build docker/ -t $(docker_image):$(tag)
+
+docker_iso: 
+	@docker run --rm $(docker_args) $(docker_image):$(tag) make iso
+
+docker_run: docker_iso
```
andre-richter: This is a duplication of the normal `run` target. I think it can be deleted?

Makefile
```diff
@@ -19,6 +19,12 @@ assembly_source_files := $(wildcard src/arch/$(arch)/*.asm)
 assembly_object_files := $(patsubst src/arch/$(arch)/%.asm, \
 	build/arch/$(arch)/%.o, $(assembly_source_files))
 
+# used by docker_* targets
+docker_image ?= blog_os
+tag ?= 0.1
+docker_args ?= -e LOCAL_UID=$(shell id -u) -e LOCAL_GID=$(shell id -g) -v blogos-$(shell id -u)-$(shell id -g)-cargo:/usr/local/cargo -v blogos-$(shell id -u)-$(shell id -g)-rustup:/usr/local/rustup -v $(shell pwd):$(shell pwd) -w $(shell pwd)
+docker_clean_args ?= blogos-$(shell id -u)-$(shell id -g)-cargo blogos-$(shell id -u)-$(shell id -g)-rustup
```
andre-richter: Putting the volume names in their own variables would avoid duplication of code in lines 25-26?

Makefile
Sh4d1: At first I thought so but when running `make run` after `make docker_iso` it did not use the iso built using Docker. I suppose there is a way to solve this I'll try to look into it, or maybe if you have some ideas :)

Makefile
Sh4d1: Right! I'll push the changes later :)

Makefile
andre-richter: Thats strange, its is an identical command, just target names differ.

phil-opp:
I decided to merge this as is. The duplication of the  `run` target can still be removed in a follow-up PR if desired. Thanks a lot for the PR, @Sh4d1! And thanks to @andre-richter for reviewing!

PR #375, opened 2017-12-15
Use custom alloc::oom handler to call panic instead of abort

robert-w-gries:
Currently, the `BumpAllocator` uses the [default `oom` handler](https://doc.rust-lang.org/1.7.0/src/alloc/oom.rs.html#17-21), which calls `intrinsics::abort()`. On the `x86_64` architecture, this generates a [`ud2` instruction](http://www.felixcloutier.com/x86/UD2.html), generating an invalid opcode exception.

In `blog_os`, allocating too much memory will call the default oom handler generate an `INVALID OPCODE` exception that points to the first `ud2` instruction in `alloc::oom` (`0x1073fc` in my case):

```asm
0000000000107ef0 <_ZN5alloc9allocator5Alloc3oom17h5474e96b9ba411a5E>:
  107ef0:	55                   	push   %rbp
  107ef1:	48 89 e5             	mov    %rsp,%rbp
  107ef4:	48 83 ec 10          	sub    $0x10,%rsp
  107ef8:	48 89 7d f8          	mov    %rdi,-0x8(%rbp)
  107efc:	0f 0b                	ud2    
  107efe:	0f 0b                	ud2    
  107f00:	48 89 75 f0          	mov    %rsi,-0x10(%rbp)
  107f04:	66 2e 0f 1f 84 00 00 	nopw   %cs:0x0(%rax,%rax,1)
  107f0b:	00 00 00 
  107f0e:	66 90                	xchg   %ax,%ax
```

I find the exception message unhelpful.  So instead of generating an exception, we can set a custom `oom` handler by implementing `Alloc` for `BumpAllocator`.  Custom `oom` handlers will be important later when we want to gracefully handle errors.  My PR results in the following message when we run out of memory.

```
PANIC in src/memory/heap_allocator.rs at line 46:
    Out of memory
```

### Testing

You can test this PR by adding the following to `src/lib.rs` and using the `BumpAllocator` instead of `LockedHeap`.

```rust
let mut stack: alloc::Vec<usize> = vec![0; 100_000_000];
```


phil-opp:
Thanks!

PR #376, opened 2017-12-15
Use proper size for heap init

robert-w-gries:
The `LockedHeap.init()` function takes `heap_bottom` and `heap_size` as arguments.  Instead, we pass `heap_bottom` and `heap_top`.

This led to unexpected page faults and other weird behavior because the hole sizes were incorrect.


phil-opp:
Good catch, seems like we messed something up during the update to the new allocator API. Thanks a lot!

Issue #379, opened 2017-12-26
Set Up Rust, Xargo compiling issues


sumberg:
Hello! I'm quite new to OS programming, taking your tutorial together with a university course. You're explanations are fantastic, thanks a lot for your hard work!

I've run in to some issues with Cargo/Xargo in the Set Up Rust part of your tutorial. I've followed your instructions and installed Cargo, then Xargo, together with all of the small necessary bits and bobs. I'm now trying to run `xargo build --target=x86_64-blog_os`, but the compiling fails. This is the output when running the same command with `-v`:

```
> xargo build --target x86_64-blog_os -v
+ "rustc" "--print" "sysroot"
+ "rustc" "--print" "target-list"
+ "cargo" "build" "--release" "--manifest-path" "/tmp/xargo.UeQjJDzmdqr9/Cargo.toml" "--target" "x86_64-blog_os" "-v" "-p" "core"
   Compiling core v0.0.0 (file:///home/usr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore)
     Running `rustc --crate-name core /home/usr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/lib.rs --crate-type lib --emit=dep-info,link -C opt-level=3 -C metadata=2574ddbde34f06b1 -C extra-filename=-2574ddbde34f06b1 --out-dir /tmp/xargo.UeQjJDzmdqr9/target/x86_64-blog_os/release/deps --target x86_64-blog_os -L dependency=/tmp/xargo.UeQjJDzmdqr9/target/x86_64-blog_os/release/deps -L dependency=/tmp/xargo.UeQjJDzmdqr9/target/release/deps --sysroot /home/usr/.xargo -Z force-unstable-if-unmarked`
error: Error loading target specification: Could not find specification for target "x86_64-blog_os"
  |
  = help: Use `--print target-list` for a list of built-in targets

error: Could not compile `core`.

Caused by:
[...]

PR #380, opened 2017-12-27
Fix travis with workaround for xargo issue

robert-w-gries:
This is a temporary CI workaround so that Travis passes until #379 is resolved


phil-opp:
Thanks a lot!

PR #383, opened 2018-01-09
Fix eh_personality reference error

bissonex:
Fix issue with undefined reference to `rust_eh_personality' when compiling in blog post


phil-opp:
Thanks! I think we also need to make it `pub` if we add `no_mangle`…

phil-opp:
Perfect, thanks!

PR #389, opened 2018-02-12
Corrections from proof read for second edition first article

andre-richter:
Some minor corrections while proof-reading.

Changed Max OS X to macOS, since it's the official Apple speak since 2016.


phil-opp:
Thanks so much!

PR #390, opened 2018-02-12
Nitpicks for second article

andre-richter:
Some nitpick corrections for #360.

I am not sure if I understood the following sentence correctly: `- The standard is written for the C programming language. That means lots of C-typical structures are present, such as linked lists in the [boot information].`

In its original form it was not really clear what is meant here. I tried to guess and correct in that way, but briefly glancing over the linked boot information, I was not able to spot linked lists :confused:

### DIFF
--- a/blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
+++ b/blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
@@ -31,7 +31,7 @@ On x86, there are two firmware standards: the “Basic Input/Output System“ (*
 Currently, we only provide BIOS support, but support for UEFI is planned, too. If you'd like to help us with this, check out the [Github issue](https://github.com/phil-opp/blog_os/issues/349).
 
 ### BIOS Boot
-Almost all x86 systems have support for BIOS booting, even newer UEFI-based machines (they include an emulated BIOS). This is great, because you can use the same boot logic across all machines from the last centuries. But this wide compatibility is at the same time the biggest disadvantage of BIOS booting, because it means that the CPU is put into a 16-bit compability mode called [real mode] before booting so that that arcane bootloaders from the 1980s would still work.
+Almost all x86 systems have support for BIOS booting, even newer UEFI-based machines (they include an emulated BIOS). This is great, because you can use the same boot logic across all machines from the last centuries. But this wide compatibility is at the same time the biggest disadvantage of BIOS booting, because it means that the CPU is put into a 16-bit compatibility mode called [real mode] before booting so that arcane bootloaders from the 1980s would still work.
 
 But let's start from the beginning:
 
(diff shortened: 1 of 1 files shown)


blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
```diff
@@ -56,14 +56,14 @@ To avoid that every operating system implements its own bootloader, which is onl
...
-- The newest version of the standard, [Multiboot 2.0], supports kernels in the 64-bit ELF format, but requires a [adjusting the default page size].
-- The development of the standard and GRUB is highly interleaved. For example, I'm not aware of an other bootloader that is Multiboot compliant. As a result, one often has to look into the GRUB source code when the standard is not clear enough.
-- The standard is written for the C programming language. That means lots of C-typical structures such as linked lists in the [boot information].
+- The newest version of the standard, [Multiboot 2.0], supports kernels in the 64-bit ELF format, but requires [adjusting the default page size].
+- The development of the standard and GRUB is highly interleaved. For example, I'm not aware of another bootloader that is Multiboot compliant. As a result, one often has to look into the GRUB source code when the standard is not clear enough.
+- The standard is written for the C programming language. That means lots of C-typical structures are present, such as linked lists in the [boot information].
```
andre-richter: In its original form it was not really clear what is meant here. I tried to guess and correct in that way, but briefly glancing over the linked boot information, I was not able to spot linked lists 😕

phil-opp:
I really should've proofread that posts a few more times. Thanks a lot!

You guessed correctly, that's what I meant. However, I'm not sure whether we want to keep that point. I will do another pass before publishing and fix or remove it then.

phil-opp:
Sorry, wrong button ^^

andre-richter:
No worries, I learn a lot from reading it, and such error are perfectly normal and just happen because you don't notice them anymore after staring at your own texts for days and days.

Regarding the linked-lists, it is a valid point if true. Its just that specifically the `Boot information format` that you linked looks like an ordinary C struct and not like a linked list, at least from just briefly skimming over it. Maybe there's more obvious places to link to for showing linked lists in multiboot?

le-jzr:
I don't know about linked lists, but to me, multiboot seems just too lenient to the bootloader. You need to handle the case where there are several small memory reservations that are not aligned to page size, and not at contiguous addresses, just fragmented at random places in the address space. It's a messy thing to deal with in your bootstrap code.

phil-opp:
@le-jzr Well put! I try to add that to the list.

> Regarding the linked-lists, it is a valid point if true. Its just that specifically the Boot information format that you linked looks like an ordinary C struct and not like a linked list, at least from just briefly skimming over it. Maybe there's more obvious places to link to for showing linked lists in multiboot?

I think I had the the memory map in mind. It's not a linked list, but a list of dynamically sized elements, so that you have to calculate the address of the next item. This is just another example for what @le-jzr said: The Multiboot standard does not define a well-defined layout, so that the bootloader can just pass whatever the BIOS outputs. The kernel then has to handle these weird formats with all of their edge cases.

PR #394, opened 2018-03-04
Add post about VGA text mode

phil-opp:
This is the adaption of the “Printing to Screen” post to the second edition.


phil-opp:
Build fails because of https://github.com/rust-lang/rust/issues/48734. Should be fixed with the next nightly.

donald-pinckney:
This will give me something to do in airports tomorrow! Should I give feedback via github line comments, or just github comments?

phil-opp:
However you like :). Thanks a lot!

blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -0,0 +1,605 @@
...
+0x6    | Brown      | 0xe                 | Yellow
+0x7    | Light Gray | 0xf                 | White
+
+Bit 4 is the _bright bit_, which turns for example blue into light blue.
+
+The VGA text buffer is accessible via [memory-mapped I/O] to the address `0xb8000`. This means that reads and writes to that address don't access the RAM, but directly the text buffer on the VGA hardware. This means that we can read and write it through normal memory operations to that address.
+
+[memory-mapped I/O]: https://en.wikipedia.org/wiki/Memory-mapped_I/O
+
+Note that memory-mapped hardware might not support all normal RAM operations. For example, a device could only support byte-wise reads and return junk when an `u64` is read. Fortunately, the text buffer supports normal reads and writes (and also bitwise operations), so that we don't have to treat it in special way.
```
donald-pinckney: Give link / citation that it is true that VGA supports normal reads and writes?

blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -0,0 +1,605 @@
...
+    buffer: &'static mut Buffer,
+}
+```
+The writer will always write to the last line and shift lines up when a line is full (or on `\n`). The `column_position` field keeps track of the current position in the last row. The current foreground and background colors are specified by `color_code` and a reference to the VGA buffer is stored in `buffer`. Note that we need an [explicit lifetime] here to tell the compiler how long the reference is valid. The [`'static`] lifetime specifies that the reference is valid for the whole program run time (which is true for the VGA text buffer).
+
+[explicit lifetime]: https://doc.rust-lang.org/book/first-edition/lifetimes.html#syntax
+[`'static`]: https://doc.rust-lang.org/book/first-edition/lifetimes.html#static
+
+### Printing
+Now we can use the `Writer` to modify the buffer's characters. First we create a method to write a single ASCII byte (it doesn't compile yet):
```
donald-pinckney: Actually it does compile

blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -0,0 +1,605 @@
...
+
+```rust
+impl Writer {
+    pub fn write_byte(&mut self, byte: u8) {
+        match byte {
+            b'\n' => self.new_line(),
+            byte => {
+                ...
+
+                self.buffer().chars[row][col].write(ScreenChar {
```
donald-pinckney: `.buffer().` should be just `.buffer.`

donald-pinckney:
Ok, I just added 3 line comments. The other general comment is: what happened to the blink field??? I was expecting to have the VGA buffer specification match precisely our code representation, but it doesn't, which feels kinda weird.

phil-opp:
Thanks a lot for the review! I addressed your comments in 2c2bd0a.

> what happened to the blink field???

My plan was to disable blinking support in the bootloader [through a BIOS call](http://www.ctyme.com/intr/rb-0117.htm). Support for bright backgrounds seems like a more sensible default than support for blinking. I haven't had the time to implement it though.

donald-pinckney:
Ok. It's not a big deal either way. As usual it was a great post.

PR #395, opened 2018-03-06
LLD is now shipped with Rust!

phil-opp:
https://github.com/rust-lang/rust/pull/48125 was merged and is now shipped in nightly :tada:!

Unfortunately it doesn't work on macOS right now: https://github.com/rust-lang/rust/issues/48772, but I hope that it's resolved soon.


phil-opp:
Merging even though the test on macOS fails right now, so that at least the Linux and Windows people can try this.

PR #401, opened 2018-03-09
Add a “Writing an OS in pure Rust” news post

phil-opp:


phil-opp:
Doesn't change any code, so we don't need to wait for travis/appveyor.

PR #402, opened 2018-03-09
Switch (possible) typo, arcane --> archaic

Redrield:
I'm fairly sure that the idea trying to be put across here is that old bootloaders can still use the code. I'm fairly sure this should be "archaic" in that case, not "arcane".


phil-opp:
Thanks!

PR #404, opened 2018-03-10
Wrong method referenced in prose

mitchblevins:
Just a typo, it seems.


phil-opp:
I think it I meant that the `write_str` method of the `core::fmt::Write` trait is similar to the `Writer::write_str` method we already have. So it's not a typo, but I see how it can seem confusing. We should make that difference more explicit in the sentence.

phil-opp:
I just merged #417, I hope this makes things a little bit clearer.

Thanks for the PR, it helped by showing a source confusion!

PR #407, opened 2018-03-10
Remove extra "the"

lambda:
There is an extra word in "the entry point is _the_ called `main`".

It's possible that "the entry point is _then_ called `main`" was intended,
but it works without the word just as well, so just remove it.


phil-opp:
Thanks!

PR #415, opened 2018-03-12
Added more specific instructions for nightly

Immortalin:
@phil-opp might want to consider putting the `-Z pre-linker` flags in cargo.toml instead of passing it in manually


phil-opp:
Thanks for the PR, @Immortalin! Not everyone wants to use nightly by default :).

bors r+

> might want to consider putting the -Z pre-linker flags in cargo.toml instead of passing it in manually

I didn't know this was possible. However, we no longer need it in the following posts, so using the command line flag makes more sense to me in this case.

Immortalin:
@phil-opp merge and close?

phil-opp:
Seems like Travis timed out. Let's try again:

bors r+

phil-opp:
Hmm, looks like bors still fails. Merging manually.

Sorry for the delay @Immortalin!

PR #421, opened 2018-03-22
Fix cargo install-update invocation

phil-opp:
See https://github.com/nabijaczleweli/cargo-update/issues/71


phil-opp:
This seems to be an issue in rustc/cargo: https://github.com/rust-lang/cargo/issues/5212. Should be fixed with the next nightly.

phil-opp:
Works again on current nightly

PR #425, opened 2018-04-02
Handle non-printable characters in vga buffer module

phil-opp:
Suggested by @fpsiorz in https://github.com/phil-opp/blog_os/issues/405#issuecomment-377820056.

I decided to modify `write_string` instead of `write_byte`, because it allows us to still print the additional bytes of codepage 437 via `write_byte` (e.g. a `■` for non-ASCII characters).

A better solution would be to translate UTF-8 to codepage 437 where possible, but I think that would be too much code for the blog. Maybe we could do it via an external crate.


blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -203,12 +203,23 @@ To print whole strings, we can convert them to bytes and print them one-by-one:
 impl Writer {
     pub fn write_string(&mut self, s: &str) {
         for byte in s.bytes() {
-            self.write_byte(byte)
+            match byte {
```
emmanuelantony2000, line 206: Instead of a `match` we could use an `if` condition:

```rust
if byte.is_ascii() {
    self.write_byte(byte);
} else {
    self.write_byte(0xfe);
}
```

bjorn3: That also includes non-printable ASCII bytes.

PR #428, opened 2018-04-25
Create a new post about unit testing

phil-opp:
This is a small post about unit testing in `no_std` applications. It does not contain much operating system related content, but I think that a working `cargo test` is useful even in kernels. The post is also a good introduction for the next planned post, which will try to create a small integration test framework that runs tests inside QEMU.

[Rendered Preview](https://github.com/phil-opp/blog_os/blob/unit-tests/blog/content/second-edition/posts/04-unit-testing/index.md)

Feedback appreciated!

### DIFF
--- a/blog/content/second-edition/posts/04-unit-testing/index.md
+++ b/blog/content/second-edition/posts/04-unit-testing/index.md
@@ -1,0 +1,397 @@
++++
+title = "Unit Testing"
+order = 4
+path = "unit-testing"
+date  = 2018-04-14
+template = "second-edition/page.html"
++++
+
+This post explores unit testing in `no_std` executables using Rust's built-in test framework. We will adjust our code so that `cargo test` works and add some basic unit tests to our VGA buffer module.
+
+<!-- more -->
+
+This blog is openly developed on [Github]. If you have any problems or questions, please open an issue there. You can also leave comments [at the bottom].
+
+[Github]: https://github.com/phil-opp/blog_os
+[at the bottom]: #comments
+
+## Unit Tests for `no_std` Binaries
+Rust has a [built-in test framework] that is capable of running unit tests without the need to set anything up. Just create a function that checks some results through assertions and add the `#[test]` attribute to the function header. Then `cargo test` will automatically find and execute all test functions of your crate.
+
+[built-in test framework]: https://doc.rust-lang.org/book/second-edition/ch11-00-testing.html
+
+Unfortunately it's a bit more complicated for `no_std` applications such as our kernel. If we run `cargo test` (without adding any test yet), we get the following error:
+
+```
+> cargo test
+   Compiling blog_os v0.2.0 (file:///home/philipp/Documents/blog_os)
+error[E0152]: duplicate lang item found: `panic_fmt`.
+  --> src/main.rs:26:1
+   |
+26 | / pub extern "C" fn rust_begin_panic(
+27 | |     _msg: core::fmt::Arguments,
+28 | |     _file: &'static str,
+29 | |     _line: u32,
+...  |
+32 | |     loop {}
+33 | | }
+   | |_^
+   |
+   = note: first defined in crate `std`.
+```
+
+The problem is that unit tests are built for the host machine, with the `std` library included. This makes sense because they should be able to run as a normal application on the host operating system. Since the standard library has it's own implementation of the `panic_fmt` language item, we get the above error. To fix it, we use [conditional compilation] to include our implementation of the language item only in non-test environments:
+
+[conditional compilation]: https://doc.rust-lang.org/reference/attributes.html#conditional-compilation
+
+
+```rust
+// in src/main.rs
+
+#[cfg(not(test))] // only compile when the test flag is not set
+#[lang = "panic_fmt"]
+#[no_mangle]
+pub extern "C" fn rust_begin_panic(
+    _msg: core::fmt::Arguments,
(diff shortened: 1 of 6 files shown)


IsaacWoods:
This is so cool! I didn't realise it was even possible to test normally, and was planning to create my own test harness around [utest](https://github.com/japaric/utest). This works as described on one of my crates in my kernel, so that's nice, but is less successful on another. I haven't done much digging, so I'm not sure if this'll actually apply to anyone else trying this, but just thought I'd mention it.

<details>
Looks like I'm executing an illegal instruction somewhere, but the only test in the crate is an empty one:

``` rust
#[test]
fn test()
{
}
```

```
thread panicked while processing panic. aborting.
error: process didn't exit successfully: `/home/isaac/Documents/pebble/kernel/x86_64/target/debug/dep
s/x86_64-c14c325372397a7d` (signal: 4, SIGILL: illegal instruction)
Thread 2 "cargo" received signal SIGUSR1, User defined signal 1.
[Switching to Thread 0x7ffff07ff700 (LWP 2872)]
0x00005555556f91d2 in core::ptr::drop_in_place ()
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.26-27.fc27.x86_64 libgcc-7.3.1-5.fc27
.x86_64
(gdb) bt
#0  0x00005555556f91d2 in core::ptr::drop_in_place ()
#1  0x000055555594c643 in std::sys_common::backtrace::__rust_begin_short_backtrace ()
#2  0x000055555594c9ce in std::panicking::try::do_call ()
#3  0x0000555555cafc9f in __rust_maybe_catch_panic () at libpanic_unwind/lib.rs:105
#4  0x000055555594df00 in <F as alloc::boxed::FnBox<A>>::call_box ()
#5  0x0000555555ca3bb8 in _$LT$alloc..boxed..Box$LT$alloc..boxed..FnBox$LT$A$C$$u20$Output$u3d$R$GT$$
u20$$u2b$$u20$$u27$a$GT$$u20$as$u20$core..ops..function..FnOnce$LT$A$GT$$GT$::call_once::h0137220ca48
23a86 () at /checkout/src/liballoc/boxed.rs:647
#6  std::sys_common::thread::start_thread () at libstd/sys_common/thread.rs:24
#7  0x0000555555c8c5e9 in std::sys::unix::thread::Thread::new::thread_start ()
    at libstd/sys/unix/thread.rs:90
#8  0x00007ffff77b550b in start_thread () from /lib64/libpthread.so.0
#9  0x00007ffff72d616f in clone () from /lib64/libc.so.6
```

</details>

blog/content/second-edition/posts/04-unit-testing/index.md
```diff
@@ -0,0 +1,397 @@
...
+error[E0277]: the trait bound `volatile::Volatile<vga_buffer::ScreenChar>: core::marker::Copy` is not satisfied
+   --> src/vga_buffer.rs:186:21
+    |
+186 |             chars: [[Volatile::new(empty_char); BUFFER_WIDTH]; BUFFER_HEIGHT],
+    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::marker::Copy` is not implemented for `volatile::Volatile<vga_buffer::ScreenChar>`
+    |
+    = note: the `Copy` trait is required because the repeated element will be copied
+```
+
+The problem is that array construction in Rust requires that the contained type is [`Copy`]. The `ScreenChar` is `Copy`, but the `Volatile` wrapper is not. There is currently no easy way to circumvent this without usinge [`unsafe`], but fortunatly there is the [`array_init`] crate that provides a safe interface for such operations.
```
robert-w-gries:
```diff
-There is currently no easy way to circumvent this without usinge [`unsafe`], but fortunatly there is the [`array_init`] crate that provides a safe interface for such operations.
+There is currently no easy way to circumvent this without using [`unsafe`], but fortunately there is the [`array_init`] crate that provides a safe interface for such operations.

phil-opp:
> Looks like I'm executing an illegal instruction somewhere, but the only test in the crate is an empty one:

Strange.. Maybe you have a .cargo/config file for that crate that changes some build settings and breaks the build?

phil-opp:
> It seems that the built-in testing framework has a limitation where the tested code is either architecture independent or the host architecture matches the target architecture?

Yes, this should be only used for architecture independent parts of the code. For target specific code you need to somehow test on the target architecture.

> I'm excited to see what you've done with integration testing!

_Have done_ is the wrong tense :D. I haven't implemented anything yet. My idea was to do something similar to utest, but less interleaved with rustc and Linux and much more basic.

You would have a set of different `_start` functions (e.g. as cargo examples), each with specified expected output. Instead of printing to the VGA buffer, the functions would print to serial output, which is mapped to stdout using QEMU. The test runner then just compares stdout with the expected output for each `_start` function and fails if there are differences. Endless loops or deadlocks can be caught with a timeout.

Or alternatively, the functions would perform the checks themselves and report a panic over the serial interface otherwise. The test runner then checks that no panics occur.

Either way, my plan is to integrate the test runner into bootimage, so that you can just do `bootimage test`. There will be some time to experiment with it before I write the next post.

blog/content/second-edition/posts/04-unit-testing/index.md
phil-opp: Thanks, fixed.

PR #433, opened 2018-05-26
Fixed an erroneous call to unwrap() in second edition vga buffer chapter

ccesare:
In an example code block in the second edition chapter on VGA buffers, there was a call to unwrap() chained with write_str(). But write_str() doesn't return anything, and thus there is nothing to unwrap().

### DIFF
--- a/blog/content/second-edition/posts/03-vga-text-buffer/index.md
+++ b/blog/content/second-edition/posts/03-vga-text-buffer/index.md
@@ -347,7 +347,7 @@ pub fn print_something() {
     };
 
     writer.write_byte(b'H');
-    writer.write_str("ello! ").unwrap();
+    writer.write_str("ello! ");
     write!(writer, "The numbers are {} and {}", 42, 1.0/3.0).unwrap();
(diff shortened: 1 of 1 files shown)


phil-opp:
I think this `unwrap` is actually correct. The `write_string` method does not return anything, but the implementation of the `write_str` method of the `fmt::Writer` trait returns a `fmt::Result`. See the [“Formatting Macros”](https://os.phil-opp.com/vga-text-mode/#formatting-macros) section on the blog:

> […] we need to implement the `core::fmt::Write` trait. The only required method of this trait is `write_str` that looks quite similar to our `write_string` method, **just with a fmt::Result return type**

I hope this clears things up!

I agree that the similar methods are confusing, but I don't know how to solve it in a better way. Let me know if you have any ideas!

ccesare:
Yeah, I think I made a mistake of confusing `write_string` with `write_str`. Sorry!

phil-opp:
No worries! Thank you for trying to fix mistakes!

PR #444, opened 2018-06-22
Fix confusing wording

Mackiovello:
I found the "even" in this sentence to be confusing, so I changed it and modified the end of the sentence to avoid redundancy.


phil-opp:
Thank you!

PR #448, opened 2018-07-02
Fix: should be `main.rs` instead of `lib.rs`

slin31:
The feature flag for `abi_x86_interrupt` needs to be in `main.rs`.
Putting the feature flag in my lib.rs did not resolve the error. I assume this was an oversight when updating this chapter from the first edition, because the error message in the blog also refers to a separate module `interrupts.rs` that was never mentioned in this edition.


phil-opp:
Thank you! Yes, seems like I forgot to update this for the second edition.

Could you also fix the error message that mentions the old `interrupts.rs` file?

phil-opp:
Thanks a lot!

PR #462, opened 2018-08-12
[second edition: 03 vga text buffer] Fix example code

kazu9su:
Writer.write_str does not return Result so it should be error if we call unwrap()

### DIFF
--- a/blog/content/second-edition/posts/03-vga-text-buffer/index.md
+++ b/blog/content/second-edition/posts/03-vga-text-buffer/index.md
@@ -347,7 +347,7 @@ pub fn print_something() {
     };
 
     writer.write_byte(b'H');
-    writer.write_str("ello! ").unwrap();
+    writer.write_string("ello! ");
     write!(writer, "The numbers are {} and {}", 42, 1.0/3.0).unwrap();
 }
 ```


blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -347,7 +347,7 @@ pub fn print_something() {
     };
 
     writer.write_byte(b'H');
-    writer.write_str("ello! ").unwrap();
+    writer.write_str("ello! ");
```
bemeurer: If you look on [line 330](https://github.com/phil-opp/blog_os/pull/462/files#diff-54c32378013c8f97b435c52c00b91d29R330), or on the docs for [fmt::Write](https://doc.rust-lang.org/std/fmt/trait.Write.html#tymethod.write_str), you'll see that it does, indeed, return a `Result`. The code is correct as-is.

phil-opp:
Thanks for the pull request @kazu9su!

There are two distinct string writing methods:

- `write_string` is a method of the writer and does _not_ return a result.
- `write_str` is a method of the `fmt::Write` trait that does return a `fmt::Result` (because the trait is defined that way). See the [Formatting Macros](https://os.phil-opp.com/vga-text-mode/#formatting-macros) section of the post for the `write_str` implementation.

So both `writer.write_str(…).unwrap()` and `writer.write_string(…)` are correct, but `writer.write_str(…)` is not (note the different method names).

How about changing it to `write_string(…)`? I think this would be less confusing than `write_str` because the initial version of `print_something` uses `write_string` too. We could also remove an unwrap, which is always a good idea.

@kazu9su What do you think about changing the line to `writer.write_string("ello! ");`?

kazu9su:
@bemeurer @phil-opp 
Thanks to describe the problem.
I clearly understand what you mentioned and what I misunderstood, so close this PR.

Thank you!

blog/content/second-edition/posts/03-vga-text-buffer/index.md
kazu9su: @bemeurer Thanks to comment on my PR. I'm sorry but I mentioned about the case of using `writer.write_string` method. In the case of using `writer.write_str`, your comment is absolutely true.

phil-opp:
Yes, `unwrap` is only necessary with `write_str`, not with `write_string`.

kazu9su:
I updated function name to `write_string`.

phil-opp:
Thanks!

Issue #465, opened 2018-09-23
Rename panic_implementation attribute to panic_handler


TuringDisciple:
The panic_implementation attribute has been deprecated for [RFC 2070](https://github.com/rust-lang/rust/issues/44489#issuecomment-416645946) and renamed to [panic_handler](https://github.com/rust-lang/rust/issues/44489#issuecomment-415140224)

Ben-PH:
Forking now. I ran into this and didn't think to put the issue up. I'm getting a lot of value out of this, will feel good to contribute.

Ben-PH:
found the commit in the rust repo

[deprecating panic_implementation](https://github.com/rust-lang/rust/tree/master/src/test/ui/panic-implementation)

phil-opp:
Fixed in #467. Thanks for reporting @TuringDisciple!

PR #467, opened 2018-09-27
Rfc 2070 panic implementation deprecated

Ben-PH:
#465 

I haven't tried compiling or running this. I'm new to this, and the changes seem trivial.
I did a grep search of  "panic_implementation" and made the changes. Mine now only comes up with 3 hits from `grep -r "panic_implementation"` which is a discusion on RFC 2070 and changes to `panic_handler`

Thank you for putting this blog together!


phil-opp:
Thanks a lot!

PR #468, opened 2018-10-01
Pub and no_mangle are no longer required for panic_handler

phil-opp:


phil-opp:
The appveyor failure is not related. Seems like there is some problem with cargo install-update…

PR #475, opened 2018-10-17
Refactor exception code into new `interrupts` module

acheronfail:
As discussed in #473 and https://github.com/acheronfail/blog_os/pull/1/files.

This PR re-writes the 6th and 7th posts so that all exception related code lives in `src/interrupts.rs`.

PR #476, opened 2018-10-17
refactor interrupts for #475

acheronfail:
This PR refactors the `hardware-interrupts` branch so it will match the re-write of the earlier posts from #475.


phil-opp:
Thank you!

PR #482, opened 2018-10-22
Minor code adjustments.

c-edw:
This PR addresses [this comment](https://github.com/phil-opp/blog_os/issues/405#issuecomment-431896303). It turns out that I had made a mistake, and the original code uses `...` instead of `..`. It's unclear which syntax out of `...` and `..=` will become standard in future, so we can leave it for now.

Thanks again.


phil-opp:
Thanks!

> It's unclear which syntax out of `...` and `..=` will become standard in future, so we can leave it for now.

I think `..=` will become standard, but I'm fine with leaving it as `...` until it becomes deprecated.

phil-opp:
Merging manually, since Github seems to still have problems with generating webhooks.

PR #492, opened 2018-10-28
Modify newline to clone instead of iterate.

c-edw:
I've been trying to see what I can do to improve the performance of the VGA text buffer code. This change removes an iterator and clones the next line instead.
As discussed, we'll be able to do a bit more if a VolatileArray is implemented. I also tried to "pre-build" an array for the `clear_row` function to improve performance, but didn't manage to do it without using Vecs.

### DIFF
--- a/blog/content/second-edition/posts/03-vga-text-buffer/index.md
+++ b/blog/content/second-edition/posts/03-vga-text-buffer/index.md
@@ -363,10 +363,7 @@ Right now, we just ignore newlines and characters that don't fit into the line a
 impl Writer {
     fn new_line(&mut self) {
         for row in 1..BUFFER_HEIGHT {
-            for col in 0..BUFFER_WIDTH {
-                let character = self.buffer.chars[row][col].read();
-                self.buffer.chars[row - 1][col].write(character);
-            }
+            self.buffer.chars[row - 1] = self.buffer.chars[row].clone();
         }
         self.clear_row(BUFFER_HEIGHT - 1);
         self.column_position = 0;
--- a/src/vga_buffer.rs
+++ b/src/vga_buffer.rs
@@ -119,10 +119,7 @@ impl Writer {
     /// Shifts all lines one line up and clears the last row.
     fn new_line(&mut self) {
         for row in 1..BUFFER_HEIGHT {
-            for col in 0..BUFFER_WIDTH {
-                let character = self.buffer.chars[row][col].read();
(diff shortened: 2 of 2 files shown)


phil-opp:
Thanks for the PR! I like this change because it makes things simpler. However, I'm not sure if this is still correct. The problem is that we're not doing a volatile write when we write the lines, because we're overwriting the complete `Volatile` type. So the compiler could in theory do some unwanted optimizations (e.g. delay or omit the write).

You say that you try to improve performance. Do you have any performance problems? Because the compiler normaly compiles iterators down to very efficient code.

c-edw:
I will close this since it defeats the purpose of using Volatiles if it can be optimised away. I would like to see what else could be done to optimise this code in future when we have const fns. Thanks!

phil-opp:
Sorry for the delay. 



> I'm not really clear on what circumstances the compiler would optimise away operations like this and why

For example the compiler could decide to delay a write to the buffer until the same address is read at some point later, because it thinks it is faster this way. This would be still correct from the compiler's viewpoint because the observable effects are the same. Only when doing a volatile write, the compiler knows that this write operations has other side effects (in our case that we see things on the screen), so it has to do the write operation exactly where it is written. (Note that this is all theoretical and I never seen the compiler doing an optimization like this, but it could be added in the future.)

> Even though the compiler likely optimises the iteration very well, the difference between an iterator and a row copy is noticable.

You're completely right. The problem is not so much the optimization capabilities of the compiler, but rather that volatile writes need to be executed exactly like they're written. So when iterating, we tell the compiler to do 80 volatile reads and 80 volatile writes per line. The compiler can't combine them into fewer larger writes, because they're volatile and must not be delayed. Which makes sense somehow, because combining them would really lead to a different observable output (no horizontal flickering). The problem is that we want this different output.

So what can we do about this? I can think of three possibilities:

- Wait for const generics and then implement a `VolatileArray` type, which allows volatile reads/writes of both individual cells and whole lines.
- Don't use the `Volatile` wrapper and instead try to always use `ptr:volatile_{read, write}` instead of normal read/writes. (Maybe introduce safe volatile read/write functions for reference types.)
- Make the `Buffer` type a normal stack allocated array of `ScreenChars` and create a separate `flush` method that writes the complete buffer to `0xb8000` in a single volatile operation. The drawback is that we need to overwrite the whole buffer on any single change, but the compiler could optimize the individual operations much better.

What do you think?

PR #495, opened 2018-11-04
Fix minor typo on cpu-exceptions page

Pear0:
I noticed a minor typo while following along with the blog on the CPU Exceptions page.


phil-opp:
Thanks!

PR #496, opened 2018-11-07
Fix minor url typo for System V ABI on cpu-exceptions page

noahrinehart:
I noticed that the link for the System V ABI was broken, and it was just due to a "-" instead of an "_". This pr should fix that.

I also want to say thanks so much for this project! It's incredibly inspiring.


phil-opp:
Thank you!

PR #498, opened 2018-11-17
Use pc-keyboard crate for translating scancodes

phil-opp:
This PR extends the "Hardware Interrupts" post to use the [`pc-keyboard`](https://docs.rs/pc-keyboard/0.3.1/pc_keyboard/index.html) crate for translating scancodes. This enables us to use all keys and not only 0-9 like before.

Thanks to @ZerothLaw for implementing scancode 1 support in `pc-keyboard` and of course @thejpster for writing the crate.


phil-opp:
@acheronfail How did you manage to get such small GIFs? I tried to add another GIF but it is more than 10 times larger than yours (after compression).

acheronfail:
@phil-opp I just use [`byzanz`](https://github.com/GNOME/byzanz) which has nice default compression, and usually always results in nice and small GIFs.

You can just `apt install byzanz` and then its usage is like:

```bash
byzanz-record -x 10 -y 10 -w 640 -h 480 ~/Desktop/out.gi
```

Granted, it's not a fancy GUI app, but it works well and gets the job done.

PR #501, opened 2018-11-18
Update to the 2018 edition of Rust

### LINKED ISSUE #499: Remove remaining uses of #[macro_use]
Continuing the discussion from https://github.com/phil-opp/blog_os/issues/405#issuecomment-437739645.

To summarize the issue: We're still using the `#[macro_use]` attribute for importing internal macros (`print`/`println`, `serial_print`/`serial_println`). The other macro imports are already replaced with normal use statements, which is the default since later Rust versions.

It seems however that local macros are not supported by the new import logic, only extern macros. I don't know the reason for this, but it seems like a strange decision. Interestingly, exported macros behave completely different: They are part of the root namespace and can be imported with normal use statements. This means that we have to do `use crate::println` and not `use crate::vga_buffer::println`, which I find very confusing (especially the "can't find …" error message).

Another issue is that we always need to import both `print` and `println`, because `println` uses `print` internally. I don't know if there is a way to avoid this.

Finally, I was not able to find any documentation that describes the new macro import system and its limitations.

cc @kballard

### LINKED ISSUE #500: Transition the code to Rust 2018
Rust 2018 is about to be released and it's already the default for new crates on nightly. So we should update this crate and the blog posts to the 2018 edition of Rust.

After transitioning:

- [ ] Use `#![cfg_attr(not(test), no_std)]` instead of `#[cfg(test)] extern crate std` as proposed [here](https://github.com/phil-opp/blog_os/issues/429#issuecomment-437750231).

phil-opp:
This updates the blog to use the upcoming [2018 edition](https://rust-lang-nursery.github.io/edition-guide/rust-2018/index.html) of Rust, which is currently in beta and already the default on nightly. It changes a few local import paths (now with a `crate::` prefix) and removes all `extern crate` definitions and `macro_use` attributes.

This PR changes a lot of code across all posts, so there might be some things in the posts that I forgot to update. Please let me know if you see anything! [Preview of the changes](https://rust-2018--blog-os.netlify.com/)

Fixes #499 
Fixes #500

### DIFF
--- a/blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
+++ b/blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
@@ -332,12 +332,6 @@ Instead of writing our own bootloader, which is a project on its own, we use the
 bootloader = "0.3.4"
 ```
 
-```rust
-// in main.rs
-
-extern crate bootloader;
-```
-
 Adding the bootloader as dependency is not enough to actually create a bootable disk image. The problem is that we need to combine the bootloader with the kernel after it has been compiled, but cargo has no support for additional build steps after successful compilation (see [this issue][post-build script] for more information).
 
 [post-build script]: https://github.com/rust-lang/cargo/issues/545
--- a/blog/content/second-edition/posts/03-vga-text-buffer/index.md
+++ b/blog/content/second-edition/posts/03-vga-text-buffer/index.md
@@ -577,25 +577,29 @@ Now that we have a global writer, we can add a `println` macro that can be used 
 [`println!` macro]: https://doc.rust-lang.org/nightly/std/macro.println!.html
 
 ```rust
+#[macro_export]
 macro_rules! println {
     () => (print!("\n"));
-    ($fmt:expr) => (print!(concat!($fmt, "\n")));
-    ($fmt:expr, $($arg:tt)*) => (print!(concat!($fmt, "\n"), $($arg)*));
+    ($($arg:tt)*) => (print!("{}\n", format_args!($($arg)*)));
 }
 ```
-Macros are defined through one or more rules, which are similar to `match` arms. The `println` macro has three rules: The first rule for is invocations without arguments (e.g `println!()`), the second rule is for invocations with a single argument (e.g. `println!("Hello")`) and the third rule is for invocations with additional parameters (e.g. `println!("{}{}", 4, 2)`).
 
-First line just prints a `\n` symbol which literally means "don't print anything, just break the line".
-Last two rules simply append a newline character (`\n`) to the format string and then invoke the [`print!` macro], which is defined as:
+Macros are defined through one or more rules, which are similar to `match` arms. The `println` macro has two rules: The first rule for is invocations without arguments (e.g `println!()`), which is expanded to `print!("\n)` and thus just prints a newline. the second rule is for invocations with parameters such as `println!("Hello")` or `println!("Number: {}", 4)`. It is also expanded to an invokation of the `print!` macro, passing all arguments and an additional newline `\n` at the end.
+
+The `#[macro_export]` attribute makes the available to the whole crate (not just the module it is defined) and external crates. It also places the macro at the crate root, which means that we have to import the macro through `use std::println` instead of `std::macros::println`.
+
+The [`print!` macro] is defined as:
 
 [`print!` macro]: https://doc.rust-lang.org/nightly/std/macro.print!.html
 
 ```rust
+#[macro_export]
 macro_rules! print {
     ($($arg:tt)*) => ($crate::io::_print(format_args!($($arg)*)));
 }
 ```
-The macro expands to a call of the [`_print` function] in the `io` module. The [`$crate` variable] ensures that the macro also works from outside the `std` crate. For example, it expands to `::std` when it's used in other crates.
+
+The macro expands to a call of the [`_print` function] in the `io` module. The [`$crate` variable] ensures that the macro also works from outside the `std` crate by expanding to `std` when it's used in other crates.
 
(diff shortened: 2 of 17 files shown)


phil-opp:
cc @kballard

blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
```diff
@@ -332,12 +332,6 @@ Instead of writing our own bootloader, which is a project on its own, we use the
 bootloader = "0.3.4"
 ```
 
-```rust
-// in main.rs
-
-extern crate bootloader;
```
kballard: Oh excellent, I forgot Rust 2018 let you do this!

blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -577,25 +558,29 @@ Now that we have a global writer, we can add a `println` macro that can be used
...
+    ($($arg:tt)*) => (print!("{}\n", format_args!($($arg)*)));
 }
 ```
-Macros are defined through one or more rules, which are similar to `match` arms. The `println` macro has three rules: The first rule for is invocations without arguments (e.g `println!()`), the second rule is for invocations with a single argument (e.g. `println!("Hello")`) and the third rule is for invocations with additional parameters (e.g. `println!("{}{}", 4, 2)`).
 
-First line just prints a `\n` symbol which literally means "don't print anything, just break the line".
-Last two rules simply append a newline character (`\n`) to the format string and then invoke the [`print!` macro], which is defined as:
+Macros are defined through one or more rules, which are similar to `match` arms. The `println` macro has two rules: The first rule for is invocations without arguments (e.g `println!()`), which is expanded to `print!("\n)` and thus just prints a newline. the second rule is for invocations with parameters such as `println!("Hello")` or `println!("Number: {}", 4)`. It is also expanded to an invokation of the `print!` macro, passing all arguments and an additional newline `\n` at the end.
```
kballard: Typo: "invokation"

blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -604,36 +589,41 @@ The [`format_args` macro] builds a [fmt::Arguments] type from the passed argumen
...
+    ($($arg:tt)*) => ($crate::vga_buffer::_print(format_args!($($arg)*)));
 }
 
+#[macro_export]
 macro_rules! println {
-    () => (print!("\n"));
-    ($fmt:expr) => (print!(concat!($fmt, "\n")));
-    ($fmt:expr, $($arg:tt)*) => (print!(concat!($fmt, "\n"), $($arg)*));
+    () => ($crate::print!("\n"));
+    ($($arg:tt)*) => ($crate::print!("{}\n", format_args!($($arg)*)));
```
kballard: I didn't realize the output of `format_args!()` could be printed like this, but it seems it can. Very cool. 👍 

I realize you quoted this trick above as the implementation of `std::println!()`, except it isn't really, `std::println!()` calls into `$crate::io::_print` directly, but what you have here actually seems more straightforward so I can see why you're doing it this way (given that `std::println!()` depends on a private `format_args_nl!()` compiler builtin).

blog/content/second-edition/posts/05-integration-tests/index.md
```diff
@@ -103,23 +97,25 @@ Like with the [VGA text buffer][vga lazy-static], we use `lazy_static` and a spi
...
+pub fn _print(args: ::core::fmt::Arguments) {
     use core::fmt::Write;
     SERIAL1.lock().write_fmt(args).expect("Printing to serial failed");
 }
 
 /// Prints to the host through the serial interface.
+#[macro_export]
 macro_rules! serial_print {
     ($($arg:tt)*) => {
         $crate::serial::print(format_args!($($arg)*));
```
kballard: This line needs to be updated to call `_print`.

blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -604,36 +589,41 @@ The [`format_args` macro] builds a [fmt::Arguments] type from the passed argumen
...
 }
 
-pub fn print(args: fmt::Arguments) {
+pub fn _print(args: fmt::Arguments) {
```
kballard: Maybe add `#[doc(hidden)]` like `std::io::_print` uses? I don't know if you want to bring up the whole question of documentation though. I feel like you can put `#[doc(hidden)]` on here without explaining it, as your readers can look that up themselves if they want to.

blog/content/second-edition/posts/05-integration-tests/index.md
```diff
@@ -103,23 +97,25 @@ Like with the [VGA text buffer][vga lazy-static], we use `lazy_static` and a spi
 To make the serial port easily usable, we add `serial_print!` and `serial_println!` macros:
 
 ```rust
-pub fn print(args: ::core::fmt::Arguments) {
+pub fn _print(args: ::core::fmt::Arguments) {
```
kballard: If you add `#[doc(hidden)]` to `vga_buffer::_print` then you'll want to add it here too.

blog/content/second-edition/posts/03-vga-text-buffer/index.md
phil-opp: Yeah, I decided against the current `std::println` implementation because of the internal `format_args_nl` macro. This is an older implementation from ~4 months ago: https://github.com/rust-lang/rust/blob/a47653214f8f8561196acf25b8898e7148f1c052/src/libstd/macros.rs#L156-L159

I thought about mentioning that we use an older version and why, but given that some content of the blog is >4 months old anyway this seemed like unnecessary information.

phil-opp:
@kballard Thanks a lot for the review!

blog/content/second-edition/posts/02-minimal-rust-kernel/index.md
phil-opp: Yes, it's really cool that we can remove all this boilerplate from the blog.

PR #504, opened 2018-12-09
Archive comments of first edition

phil-opp:
There were no new comments on the first edition posts after the second edition went live, so keeping the old commenting system ([isso](https://posativ.org/isso/)) around does not make sense anymore.

This pull request moves all the comments into static html files that are included by our static site generator on build time. This allows us to shut down the isso backend and remove a javascript file from the website. New comments (or votes) on the first edition are no longer possible, but this should be fine (there were no new comments since months and we explicitly mention on every page that the first edition is no longer updated).


blog/templates/first-edition/comments.html
```diff
@@ -0,0 +1,31 @@
...
+    {% elif page.path == "catching-exceptions/" %}
+        {% include "first-edition/comments/catching-exceptions.html" %}
+    {% elif page.path == "better-exception-messages/" %}
+        {% include "first-edition/comments/better-exception-messages.html" %}
+    {% elif page.path == "returning-from-exceptions/" %}
+        {% include "first-edition/comments/returning-from-exceptions.html" %}
+    {% else %}
+        No comments.
+    {% endif %}
+{% endmacro post_link %}
```
bjorn3: Did you mean `{% endmacro comment %}`?

phil-opp: Thanks! Fixed in ed062f2.

Issue #514, opened 2019-01-02
Duplicate method defined in 01-freestanding-rust-binary/index.md


insha:
The current version of the 01-freestanding-rust-binary/index.md has  the `main()` method define twice in the Summary section:

Current `index.md`:
```rust
...
// On Linux:
#[no_mangle] // don't mangle the name of this function
pub extern "C" fn _start() -> ! {
    // this function is the entry point, since the linker looks for a function
    // named `_start` by default
    loop {}
}

// On Windows:
#[no_mangle]
pub extern "C" fn mainCRTStartup() -> ! {
    main();
}

#[no_mangle]
pub extern "C" fn main() -> ! {
    loop {}
}

// On macOS:

#[no_mangle]
pub extern "C" fn main() -> ! {
    loop {}
}
```

This results in the following error when the freestanding binary is built:
```
error[E0428]: the name `main` is defined multiple times
  --> src/main.rs:34:1
   |
27 | pub extern "C" fn main() -> ! {
   | ----------------------------- previous definition of the value `main` here
...
34 | pub extern "C" fn main() -> ! {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `main` redefined here
   |
   = note: `main` must be defined only once in the value namespace of this module

error: aborting due to previous error
```

The first definition of the `main()` method should be removed for the build to complete successfully.

PR #515, opened 2019-01-02
Removing duplicate method definition (#514)

insha:
This resolved the issue #514 when building the freestanding binary.


phil-opp:
Hi, thanks for the PR! I didn't intend that all three entry points should be used in the same file. If you want to do that you should use [conditional compilation](https://doc.rust-lang.org/1.30.0/book/first-edition/conditional-compilation.html).

I see how the post is a bit confusing in this regard. I opened #516 to clarify it.

insha:
No worries. I was following the example manually (only typed in what was needed for my OS) and when I got to the summary section, I wasn't sure if it was intended or not so I solved for the obvious error.

Thank you for the great tutorial! 🙇

PR #524, opened 2019-01-18
Mention `cargo test --lib`

### LINKED ISSUE #523: Run tests with cargo test --lib
Currently, unit tests are run for all targets, including integration tests (bin/*).
This messes up build log.
This issue can be solved by cargo test --lib AFAIK.

phil-opp:
Fixes #523

### DIFF
--- a/blog/content/second-edition/posts/05-integration-tests/index.md
+++ b/blog/content/second-edition/posts/05-integration-tests/index.md
@@ -410,7 +410,7 @@ fn panic(info: &PanicInfo) -> ! {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


MikailBag:
I must mention I'm not expert in Rust or OSDev.
I just decided to tell you about my solution to this problem.
Also I suggest you to use this command in .travis.yml.

phil-opp:
> I just decided to tell you about my solution to this problem.

Thanks for doing that :).

> Also I suggest you to use this command in .travis.yml.

I don't mind the additional output on travis and I think it's better to run all available tests since we might add tests to the binary (or doc tests) in the future.

PR #530, opened 2019-01-21
fix unreachable link

zyctree:
seems rustc-guide already move to https://rust-lang.github.io/rustc-guide/


phil-opp:
Thanks!

bors r+

Issue #520, opened 2019-01-14
Leverage git history to simultaneously maintain working examples for each post


Ericson2314:
Have you considered having a branch for the code for each post in the series? Each branch woudl containing all the code in the code snippets together in a working a working example, but nothing additional from later installments to distract. Crucially, when one post builds on a previous one, it would be a fork of it. That means if something in Rust unstable improves affecting multiple blog posts, you can update just manually update the first blog post effected by it, and then merge into later posts' branches to propagate the change to the rest. `master`, as the grand culmination of all the blog posts, would have most of the other branches merged into it.

This first occurred to me years ago, but now that you are embarking on a second edition, I hope might be more pertinent.

phil-opp:
I did something similar with the first edition. There were (and still are) `post_X` tags with the source code for each post. For example https://github.com/phil-opp/blog_os/tree/post_4. The history of these tags is independent of the master branch and each change that updated a post's code was cherry-picked to the corresponding tag. On each change, I rebased the tags on top of each other to keep them up to date. I used tags instead of branches because the posts are quite similar to releases, which are often marken with git tags. Also, we don't clutter the branch list this way.

I think it's worth to keep the master branch separate to avoid impacting the normal work flow. This way, changes that affect multiple posts are still possible with a single commit/pull request. Also, we can keep the blog posts and the configuration for the static site generator separate, and only need to backport code changes.

I wanted to add such tags to the second edition too, I just haven't found the time for it yet. What do you think about this approach?

Ericson2314:
So first of all, anything that involves rebasing non-upstream branches means any sort of collaboration is hard. It's possible with `rebase foo --onto bar` to rebase on your rebase, but it requires a lot of care. I once had a project where I did lots of rebasing (sort of bisecting bugs; the process was a port from Windows to Linux), and no one else could do very much.

Here's a diagram of what I was proposing.
```
  Post 1 original           Post 1 revision
       v                         v
       o-------------------------o <== post 1
        \                         \
        |\                        |\
        | \ Post 2 original       | \ Post 2 updated from Post 1
        |  \ v                    |  \ v
        |   o-------------------------o <== post 2
        |    \                    |    \
        |     \                   |     \
        |      \                  |      \
        |       \                 |       \
        |        \                |        \
        |         \               |         \
        |          \              |          \
- - ---o------------o-------------o-----------o <= master
        ^           ^             ^           ^
     Merge 1     Merge 2     Merge 1 rev     Merge 2 rev
[...]

PR #538, opened 2019-01-25
[WIP] Introduce post branches

phil-opp:
See #520


Ericson2314:
Did you decide to keep master disjoint for now?

phil-opp:
Yes, for now I just squash the latest post branch into master on each change.

For the future, I can think of the following solutions:

- Merge the post branches.This is the cleanest solution from a theoretical perspective but it extremely clutters the history (>=10 merge commits for every change affecting all posts). Using `git log` without `--no-merges` becomes unbearable.
- Keep the master branch disjoint and automate the squashing (perform merge with `--squash` and `theirs` conflict strategy, except for the Readme).
- Remove the code from the `master` branch completely and explain the post branches prominently in the Readme. We could still keep the build instructions and just add an additional `git worktree add code post-10` step to check out the code in a `code` subdirectory.

I'm leaning towards the third option, I'm just not sure about the potential confusion of users and contributors.

Ericson2314:
Sounds good to me! 3 is indeed a clean solution.

PR #540, opened 2019-01-28
Try to fix caching for Windows builds on Travis CI

phil-opp:


phil-opp:
Doesn't work. Seems like we have to wait for https://github.com/travis-ci/casher/pull/38.

PR #550, opened 2019-02-07
don't require nightly compiler from freestanding binary

nivkner:
adding linker args can be done with the stable `-Clink-arg`.

the difference with `-Zpre-link-arg` is that the args are appended instead of prepended. is that distinction important in this case?


phil-opp:
Nice, thank you!

PR #552, opened 2019-02-07
Delete code from master branch

### LINKED ISSUE #520: Leverage git history to simultaneously maintain working examples for each post

phil-opp:
The code for each post now lives in the `post-XX` branches where `XX` is the post number. This makes it possible to see the intermittent state of the code after each post. Code changes are pushed to the earliest affected `post-XX` branch and then subsequently merged into the following branches.

In theory, we could just merge the latest `post-XX` branch into `master` on each change to keep it up to date. However, this would extremely clutter the git history since they're up to ten merge commits for each change (`post-01` merged into `post-02`, `post-02` merged into `post-03` and so on). Therefore, this pull request takes a different approach.

Instead of keeping the master branch up to date, this pull request removes all code from it and instead points to the post branches from the Readme. This makes it a bit more difficult to find the code, but it has multiple advantages:

- The readers directly learn about the post branches, which are much more useful when following the blog than the old master branch.
- There is no way to accidentally update the code in the master branch instead of the post branches.
  - Thus, we never need to backport any changes from master to the post branches.
- It improves the separation of concerns: The post branches only contain the code and the master branches only contain the blog sources.
  - No CI jobs are needed for the master branch anymore (apart from the netlify build checks), so we don't need to wait for CI before merging typo fixes.

See #520 and #538 for more information.

cc @Ericson2314

Closes #520


phil-opp:
Awesome! Thanks suggesting these changes!

phil-opp:
Let's try whether bors still works:

bors r+

phil-opp:
It seems like netlify does not set commit statuses for branches, only for pull requests. I'm just removing the bors.toml now since I don't think that we need bors for master branch.

PR #557, opened 2019-02-12
encapsulate the hardware interrupts' indexes + minor corrections

AntoineSebert:
I have changed the source code but I don't know how to reflect the changes on the blog.
This is the first time I contribute to something, I don't really know what I'm doing exactly.
I hope it does not only works for me, it would be nice if someone could clone my modifications locally to be sure it's actually working.
The first commit (9f6ea55) is the most important and the changes are precisely detailed in the commit description.

### DIFF
--- a/src/interrupts.rs
+++ b/src/interrupts.rs
@@ -15,13 +15,13 @@ pub const PIC_2_OFFSET: u8 = PIC_1_OFFSET + 8;
 
 pub mod interrupt_indexes {
     #[derive(Debug)]
-	#[repr(u8)]
-	pub enum Hardware {
-		Timer = super::PIC_1_OFFSET,
-		Keyboard,
-	}
-	pub fn as_u8(interrupt: Hardware) -> u8 { interrupt as u8 }
-	pub fn as_usize(interrupt: Hardware) -> usize { usize::from(as_u8(interrupt)) }
+    #[repr(u8)]
+    pub enum Hardware {
+        Timer = super::PIC_1_OFFSET,
+        Keyboard,
+    }
+    pub fn as_u8(interrupt: Hardware) -> u8 { interrupt as u8 }
+    pub fn as_usize(interrupt: Hardware) -> usize { usize::from(as_u8(interrupt)) }
 }
 
 pub static PICS: spin::Mutex<ChainedPics> =
@@ -36,9 +36,9 @@ lazy_static! {
                 .set_handler_fn(double_fault_handler)
                 .set_stack_index(gdt::DOUBLE_FAULT_IST_INDEX);
         }
-	    use self::interrupt_indexes::*;
-	    idt[as_usize(Hardware::Timer)].set_handler_fn(timer_interrupt_handler);
-	    idt[as_usize(Hardware::Keyboard)].set_handler_fn(keyboard_interrupt_handler);
+        use self::interrupt_indexes::*;
+        idt[as_usize(Hardware::Timer)].set_handler_fn(timer_interrupt_handler);
+        idt[as_usize(Hardware::Keyboard)].set_handler_fn(keyboard_interrupt_handler);
         idt
     };
 }
--- a/src/interrupts.rs
+++ b/src/interrupts.rs
@@ -15,12 +15,12 @@ pub const PIC_2_OFFSET: u8 = PIC_1_OFFSET + 8;
 
 #[derive(Debug, Clone, Copy)]
 #[repr(u8)]
-pub enum HardwareInterruptIndexes {
+pub enum InterruptIndex {
     Timer = PIC_1_OFFSET,
     Keyboard,
 }
 
-impl HardwareInterruptIndexes {
+impl InterruptIndex {
     fn as_u8(self) -> u8 { self as u8 }
     fn as_usize(self) -> usize { usize::from(self.as_u8()) }
 }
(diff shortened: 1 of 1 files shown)


AntoineSebert:
What should I do now ?
Is everything alright ?

phil-opp:
@AntoineSebert Thanks a lot for the pull request! You're doing everything right :).

> I hope it does not only works for me, it would be nice if someone could clone my modifications locally to be sure it's actually working.

We're using various continuous integration (CI) services that automatically run tests on every pull request. You can see these checks at the bottom of this page. So you can't really break anything accidentally.

---

I think the code could be simplified a bit. Instead of creating the `interrupt_indexes` module you could add the `as_u8` and `as_usize` functions directly to the enum. I have created a small example [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d43ba8bc84d5f4d234941016ab092584). Other than that it looks good to me!

AntoineSebert:
@phil-opp I did not know such a thing was possible, that's awesome !
I have started to modify the code according to your remark.
By the way, you may have noticed the `self::` part of the `use` directives: it's because my code cannot compile without it. It's strange because you apparently don't need it. I don't know the reason yet.

phil-opp:
> By the way, you may have noticed the `self::` part of the `use` directives: it's because my code cannot compile without it. It's strange because you apparently don't need it. I don't know the reason yet.

There are two types of import paths in Rust:

- Imports of an extern crate. These start with the name of the extern crate (e.g. `x86_64::`).
- Imports from within the current crate. Paths can start from the crate root (`crate::`), from the current module (`self::`), or from the parent module (`super::`).

There is also a new feature called [uniform paths](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html#uniform-paths) in nightly Rust, which allows to omit `self::` in paths as long as they're unique. Maybe that is causing the confusion.

AntoineSebert:
@phil-opp So should I leave it like that ?
It does not seem to be an important issue, I just would like not to break consistency in your repository.
I'm quite happy that all the checks passed for the previous version of my changes, now I'm waiting for the new changes to be checked and I'm more confident about that.

src/interrupts.rs
```diff
@@ -13,8 +13,17 @@ use x86_64::structures::idt::{ExceptionStackFrame, InterruptDescriptorTable};
 pub const PIC_1_OFFSET: u8 = 32;
 pub const PIC_2_OFFSET: u8 = PIC_1_OFFSET + 8;
 
-pub const TIMER_INTERRUPT_ID: u8 = PIC_1_OFFSET;
-pub const KEYBOARD_INTERRUPT_ID: u8 = PIC_1_OFFSET + 1;
+#[derive(Debug, Clone, Copy)]
+#[repr(u8)]
+pub enum HardwareInterruptIndexes {
```
phil-opp: Could you rename this to `InterruptIndex`? It's a bit shorter and we will introduce some software interrupts in future posts for that we can also use this enum (so it's not only for hardware interrupts).

src/interrupts.rs
```diff
@@ -28,8 +37,9 @@ lazy_static! {
                 .set_handler_fn(double_fault_handler)
                 .set_stack_index(gdt::DOUBLE_FAULT_IST_INDEX);
         }
-        idt[usize::from(TIMER_INTERRUPT_ID)].set_handler_fn(timer_interrupt_handler);
-        idt[usize::from(KEYBOARD_INTERRUPT_ID)].set_handler_fn(keyboard_interrupt_handler);
+        use self::HardwareInterruptIndexes::*;
+        idt[Timer.as_usize()].set_handler_fn(timer_interrupt_handler);
+        idt[Keyboard.as_usize()].set_handler_fn(keyboard_interrupt_handler);
```
phil-opp: I think it fits the post better to just use `InterruptIndex::Timer` directly without the `use` statement because we don't introduce the handler functions at the same time in the post. (It's more straightforward to just add the keyboard line later without adding an additional import.)

src/interrupts.rs
AntoineSebert: Alright, I'm working on it.
Although I was also considering the software interrupts, hence the previous version's `mod interrupts` that could encompass both `enum` for the software and hardware interrupts.

src/interrupts.rs
AntoineSebert: I do not understand all that you say, but i'm going to modify the code.
I added `use` because I wanted that piece to stick to the DRY principle, but it's not big deal.

src/interrupts.rs
phil-opp: Using a single enum has the advantage that we can't accidentally use an index that it's already in use for a hardware interrupt (they all share the same IDT anyway).

src/interrupts.rs
phil-opp: Yeah sorry, my explanation wasn't good. What I meant is the following:

In the blog post, we first introduce the timer interrupt. An use statement does not make much sense with only one line that uses it, so we introduce it as

```rust
idt[InterruptIndex::Timer.as_usize].set_handler_fn(…)
```

Later in the post, we add the keyboard interrupt. At this point, it's simpler to add

```rust
idt[InterruptIndex::Keyboard.as_usize].set_handler_fn(…)
```

than to add an `use` statement and modify the previous timer interrupt line again.

I didn't meant to say that an use statement isn't good here, just that it would be more cumbersome to explain in the post. And the code should be as similar to the post as possible.

src/interrupts.rs
AntoineSebert: If the first `enum` for the software interrupts starts at **0** and contains *32* attributes, the last attribute will have the value **31**.
Then if the second `enum` for the hardware interrupts start at **32** and contains *16* attributes, the last attribute will have the value **47** :)
And last but not least, some people (like myself) might also want to implement their own interrupts, but creating another `enum`, for example `enum real_time`, that could start at **48**.
But that is a bit out of the scope of your tutorial, so I can keep it for myself.

src/interrupts.rs
AntoineSebert: Oh right that makes sense.
Sometimes I forgot that your repository exists for the purpose of the OS tutorial x)

AntoineSebert:
You're welcome, it was my first time contributing to something and the experience was very enjoyable.
I will keep in touch with the future evolutions of your tutorial, which has been very helpful so far.

phil-opp:
I'm glad to hear that! Thank you for your contribution!

src/interrupts.rs
phil-opp: Yeah, that's possible of course :).

PR #562, opened 2019-03-07
fix typos

AntoineSebert:
`create_example_mapping` was written `create_mapping` in #L709, #L949, #L964


phil-opp:
Ah, I just saw that you opened this PR against the paging-implementation branch instead of master. It's fine too since paging-implementation is the work in progress branch for the upcoming rewrite of the advanced paging post, but could you also fix the current post in the master branch?

AntoineSebert:
Alright, I'm on it

PR #566, opened 2019-03-09
Update to version 0.5.0 of x86_64

phil-opp:
A changelog can be found here: https://github.com/rust-osdev/x86_64/pull/61.


phil-opp:
The `post-XX` branches were updated too.

PR #579, opened 2019-04-24
Migrate from Netlify to Github Pages

phil-opp:
I recently got a mail from Netlify that I used 65% of the free 100GB bandwidth in the last month. I don't quite understand why we needed so much bandwidth and Netlify provides no finer-grained bandwidth statistics (e.g. traffic per site), so we have to trust what they say. Additional bandwidth costs 20$ per 100GB per month.

I contacted them to ask about their open source plan, which has a higher free bandwidth. Unfortunately, we don't quite meet their conditions for the open source plan because the content of the blog is not available under a free license (only the code and the used tools/libraries). I don't want to change that, at least for now.

For these reasons, this PR migrates the blog from Netlify back to GitHub Pages. They also have a 100GB bandwidth limit, but this is only a soft limit and according to the official support it shouldn't be enforced for this project.

On the technical side, this PR performs the following changes:

- Add a azure pipelines script that builds the site using zola and outputs the generated site as release artifact
  - I set up a deploy script that pushes the changes to the `gh-pages` branch of the [blog_os_deploy](https://github.com/phil-opp/blog_os_deploy) repository, where Github Pages is set up. I used a separate repository to keep the activity log of this repository clean and not spam people that watch this repository.
  - Build all branches, not just `master` and `post-*` like before
- Add a CNAME file that is used by Github Pages
- Instead of Netlify's redirect feature, use the aliases feature of Zola
  - This replaces server-side redirects with inferior meta-refresh redirects. However, most search engines handle meta-refresh redirects too, so this shouldn't be a big problem.
- Link to `/` instead of `/second-edition`, which is redirected to `/`
- Remove netlify.toml and link to netlify.com


phil-opp:
I just updated the DNS to point to GitHub Pages instead of Netlify. The DNS changes will probably need a few hours to propagate. There will also be cerficate errors until GitHub Pages set up the new Letsencrypt SSL certificate. Sorry for inconvenience!

phil-opp:
As a temporary workaround you can use https://os-old.phil-opp.com/ to access the old netlify-hosted site.

bjorn3:
> The DNS changes will probably need a few hours to propagate. There will also be cerficate errors until GitHub Pages set up the new Letsencrypt SSL certificate.

It works for me.

phil-opp:
@bjorn3 Thanks for testing! It work for me too by now.

PR #601, opened 2019-05-07
Remove unecessary `use` statement

AntoineSebert:
The line `use core::fmt::Write;` is probably not necessary since `Write` is not present.


phil-opp:
I think the import is needed because we use the `writeln!` macro, which tries to call the `write_fmt` macro of the imported trait.

PR #608, opened 2019-05-11
Trait object link points to the 2018 edition of the book

3kyro:
On chapter 04-testing, fixed a link pointing to the old version of the book.


phil-opp:
Thanks for the pull request! I deliberately linked to the first edition chapter because I think it conveys the information about trait objects quicker, as it's more reference-like and does not require you to follow an example. However, I'm a bit biased since I learned Rust with the first edition book, so I might prefer it just because it is more familiar.

Which trait object explanation do you prefer? The third alternative is the reference, which also has a [chapter about trait objects](https://doc.rust-lang.org/reference/types/trait-object.html).

3kyro:
Hi, you are right, the first edition of the book does give a more concise view of the trait object plus better explains dynamic dispatch . Since I started learning rust using the 2018 version, I automatically switch to the newer version as the red banner seems menacing :) I will close the PR.

phil-opp:
Thanks for looking into this!

PR #609, opened 2019-05-12
Always use spaces for code (if possible)

josephlr:
Almost all the code in the blog uses spaces instead of tabs.
This change fixes 3 places where there was inconsistancy.

This was causing some of the content to appear misaligned.

Now tabs are only used in:
  - Makefiles
  - Dockerfiles
  - Command output (from `objdump` and `diff`) containing tabs
  - `.fish` files


phil-opp:
Thanks!

PR #610, opened 2019-05-22
Added false to write_str

xj42:
WIthout the ability to set equals to false, the function doesn't correctly handle failing comparisions.

### DIFF
--- a/tests/panic_handler.rs
+++ b/tests/panic_handler.rs
@@ -66,6 +66,8 @@ impl fmt::Write for CompareMessage {
     fn write_str(&mut self, s: &str) -> fmt::Result {
         if s == MESSAGE {
             self.equals = true;
+        } else {
+            self.equals = false;
         }
         Ok(())
     }


phil-opp:
Thanks for the pull request! It definitely improves the situation, but it is still a hacky solution that still has its problems. For example, `panic!("x{}", MESSAGE)` still succeeds.

For this reason I decided to rewrite this function completely using a different approach that checks the whole string and not just individual string components. I will open a PR shortly.

Thanks again for reporting this issue!

phil-opp:
I opened https://github.com/phil-opp/blog_os/pull/611 with the new version of the `CompareMessage` struct.

xj42:
Glad to have been some small help.

I tried your solution and it works a treat. 

Writing a kernel is not the best way to learn rust, but it sure is fun.
Thanks for the tutorials!

phil-opp:
@xj42 Thanks for testing!

> Writing a kernel is not the best way to learn rust, but it sure is fun.
> Thanks for the tutorials!

You're welcome :). Let me know if you have any questions, I try to make the tutorials accessible to Rust newcomers too.

PR #611, opened 2019-05-22
Rewrite CompareMessage to check the whole string

phil-opp:
Instead of checking just a single string component like before.

Fixes the issues reported in https://github.com/phil-opp/blog_os/issues/591#issuecomment-494615540. Alternative to https://github.com/phil-opp/blog_os/pull/610.

cc @xj42


phil-opp:
Post update in c8d1bbeef83687092a6a8f76b9a15aba0cb90b1e.

PR #614, opened 2019-05-27
add  #![feature(exclusive_range_pattern)]

mauriciogg:
Getting the following compilation error when directive is not specified:

error[E0658]: exclusive range pattern syntax is experimental
  --> src/vga_buffer.rs:74:17
   |
74 |                 0x20..0x7e | b'\n' => self.write_byte(byte),
   |                 ^^^^^^^^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/37854
   = help: add #![feature(exclusive_range_pattern)] to the crate attributes to enable

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,7 +3,7 @@
 #![feature(custom_test_frameworks)]
 #![test_runner(crate::test_runner)]
 #![reexport_test_harness_main = "test_main"]
- #![feature(exclusive_range_pattern)]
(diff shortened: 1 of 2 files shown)


phil-opp:
Thanks for the pull request!

I think we're using an _inclusive_ range in the blog post (three dots), so that the feature flag shouldn't be needed.

mauriciogg:
Oh yeah you're right. That's a bug on my copying skills.

PR #625, opened 2019-06-26
New post about heap allocation

phil-opp:
This pull request adds a new blog post about heap allocation.

First, it gives an introduction to dynamic memory and shows how the borrow checker prevents common allocation errors. It then implements the basic allocation interface of Rust, creates a heap memory region, and sets up an allocator crate. At the end of this post all the allocation and collection types of the built-in alloc crate will be available to our kernel.

Code update in #626.


blog/content/second-edition/posts/10-heap-allocation/index.md
```diff
@@ -0,0 +1,782 @@
...
+
+#![feature(alloc_error_handler)] // at the top of the file
+
+#[alloc_error_handler]
+fn alloc_error_handler(layout: alloc::alloc::Layout) -> ! {
+    panic!("allocation error: {:?}", layout)
+}
+```
+
+The `alloc_error_handler` function is still unsafe, so we need a feature gate to enable it. The function receives a single argument: the `Layout` instance that was passed to `alloc` when the allocation failure occurred. There's nothing we can do to resolve that failure, so we just panic with a message that contains the `Layout` instance.
```
bjorn3: *unstable

phil-opp: Good catch, thanks!

blog/content/second-edition/posts/10-heap-allocation/index.md
```diff
@@ -0,0 +1,782 @@
...
+The first step is to define a virtual memory region for the heap. We can choose any virtual address range that we like, as long as it is not already used for a different memory region. Let's define it as the memory starting at address `0x_4444_4444_0000` so that we can easily recognize a heap pointer later:
+
+```rust
+// in src/allocator.rs
+
+pub const HEAP_START: usize = 0x_4444_4444_0000;
+pub const HEAP_SIZE: usize = 100 * 1024; // 100 KiB
+```
+
+We set the heap size to 1 KiB for now. If we need more space in the future, we can simply increase it.
```
bjorn3: *100kb

phil-opp: Thanks!

blog/content/second-edition/posts/10-heap-allocation/index.md
```diff
@@ -0,0 +1,782 @@
...
+        let heap_end_page = Page::containing_address(heap_end);
+        Page::range_inclusive(heap_start_page, heap_end_page)
+    };
+
+    for page in page_range {
+        let frame = frame_allocator
+            .allocate_frame()
+            .ok_or(MapToError::FrameAllocationFailed)?;
+        let flags = PageTableFlags::PRESENT | PageTableFlags::WRITABLE;
+        unsafe { mapper.map_to(page, frame, flags, frame_allocator)?.flush() };
```
bjorn3: Flushing is only required after the last map, right? That would be faster.

phil-opp: You can either flush each page individually, or flush the whole TLB after the last map. The former requires more instructions, but the latter flushes everything and thus causes TLB misses. I'm not sure what's faster.

bjorn3: I didnt know there was an instruction to flush a single page from the tlb. I believe I read somewhere that making a page that hasnt been present since the last full flush present doesnt need a flush afterwards. Because those heap pages have never been present, the `MapperFlush` can be `ignored()`.

phil-opp: The problem is that modern CPUs do all kind of speculative prefetching, including page walk results (see http://blog.stuffedcow.net/2015/08/pagewalk-coherence/). So I think it's better to do to many flushes than causing undefined behavior because of a missing flush.

PR #633, opened 2019-07-04
Automatically determine line where panic was thrown

laerling:
I think there should be as few cases as possible where one part of the code has to be changed manually when another part changes.

This PR makes the panic_handler test determine the line where the panic was thrown, all by itself.

### DIFF
--- a/tests/panic_handler.rs
+++ b/tests/panic_handler.rs
@@ -9,12 +9,14 @@ use core::{
 };
 
 const MESSAGE: &str = "Example panic message from panic_handler test";
-const PANIC_LINE: u32 = 17; // adjust this when moving the `panic!` call
+static mut PANIC_LINE: u32 = 0;
 
 #[no_mangle]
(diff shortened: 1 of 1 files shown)


laerling:
If you want I can also adjust the blog entry in the master branch and make a second PR for that.
But you can do that probably a lot faster.

laerling:
Apparently the fact that the assignment to `PANIC_LINE` and the invocation of the panic macro have to be on the same line will make cargofmt always fail.

phil-opp:
Thanks for the pull request! I agree that it's unfortunate to require manual changes to `PANIC_LINE` whenever the panic is moved. However, I'm not sure if using an unsafe mutable static and using non-standard formatting is really better in this case. The test is fairly small and probably never touched again, so I think I prefer the simpler `const` variant with manual adjustments in this case.

laerling:
Understandable. Have a nice day :)

phil-opp:
Thanks again for creating this pull request!

PR #634, opened 2019-07-04
Introduction to Paging: fix a couple of small typos

leocassarani:
I noticed two small typos while reading chapter 8.


phil-opp:
Thanks a lot!

PR #654, opened 2019-08-09
Fix typo in 04-testing

greenonion:
Thanks for the excellent work!


phil-opp:
Good catch, thanks! Could you fix the function call in the code example too (line 887)?

greenonion:
Ugh, I had fixed locally so it would work but somehow forgot to change it here. It's done now!

phil-opp:
Thanks a lot!

PR #663, opened 2019-09-11
Set up GitHub Actions for post-04 branch

phil-opp:


phil-opp:
I updated remaining post-XX branches manually because they were just a simple merge.

PR #668, opened 2019-09-15
fix "non-primitive cast: `x86_64::addr::VirtAddr` as `u64`"

AntoineSebert:
```
error[E0605]: non-primitive cast: `x86_64::addr::VirtAddr` as `u64`
  --> tests\heap_allocation.rs:25:41
   |
25 |     let mut mapper = unsafe { memory::init(phys_mem_offset as u64) };
   |                                            ^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait
```


phil-opp:
I changed the parameter type from `u64` to `VirtAddr` yesterday in https://github.com/phil-opp/blog_os/pull/666. Sorry for the confusion!

PR #669, opened 2019-09-15
Fix non-primitive cast: `x86_64::addr::VirtAddr` as `u64`

AntoineSebert:
Initial condition:
`memory::init()` expects `u64` as argument
Problem:
The type of `phys_mem_offset` is `x86_64::VirtAddr`
Solution:
Call the member function `as_u64()` to cast `phys_mem_offset` as `u64`

### DIFF
--- a/blog/content/second-edition/posts/10-heap-allocation/index.md
+++ b/blog/content/second-edition/posts/10-heap-allocation/index.md
@@ -682,7 +682,7 @@ fn main(boot_info: &'static BootInfo) -> ! {
 
     blog_os::init();
     let phys_mem_offset = VirtAddr::new(boot_info.physical_memory_offset);
-    let mut mapper = unsafe { memory::init(phys_mem_offset) };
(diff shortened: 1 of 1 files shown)


phil-opp:
I changed the parameter type of `memory::init` from `u64` to `VirtAddr` yesterday as part of #666:

> * _Code Change_: Update the functions of the `memory` module to take the `physical_memory_offset` as a `VirtAddr` instead of a plain `u64`. This makes it clearer that the address is virtual and also simpler to calculate the virtual address for a physical address (no more `Virt::new`).

See this [diff](https://github.com/phil-opp/blog_os/pull/667/files) for the full set of the changes. Sorry for the breakage!

AntoineSebert:
Sorry I missed that change !

phil-opp:
No worries! I try to communicate such changes better in the future. Thanks for taking the time to open a pull request!

PR #680, opened 2019-10-19
Added badges to Readme.md

AntoineSebert:
Badges make projects more newbie-friendly by making useful information about the project itself easily visible, although they should not be too many to keep the document clean.

The following badges have been added (those marked with "*" are updated automatically):
- Travis CI *
- Appveyor *
- Gitter *
- Code size *
- License
- Latest tag *

I am not sure whether the Appveyor badge is properly working or not. The proper link to the status badge is probably accessible through the Appveyor page of the repository , that I cannot access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,12 @@
 # Blog OS
 
+[![Build Status](https://travis-ci.org/phil-opp/blog_os.svg?branch=master)](https://travis-ci.org/phil-opp/blog_os)
+[![Build Status](https://ci.appveyor.com/api/projects/status/phil-opp/blog_os)]()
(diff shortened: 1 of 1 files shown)


AntoineSebert:
@phil-opp So do you think is it a good idea or should I close it ? :)

phil-opp:
Sorry for the late reply

I'm not sure if these badges are a good idea for the project because the master branch does not contain the code. So things like code size and build status are not really important. Further, we don't do releases so there is no use in displaying the latest tag (which points to the outdated first edition). The gitter chat and the license are already mentioned in the readme, so I don't see any advantage in the badges.

AntoineSebert:
Alright !

PR #682, opened 2019-10-25
Translate part to Chinese

TheBegining:
This is a great article, its content is not very complicated, but it is a bit time consuming for non-native English speakers. 
So I translated a small part into Chinese. 
If the author has a plan to translate into other languages, I believe that more people will contribute.

### DIFF
--- a/blog/content/second-edition/posts/01-freestanding-rust-binary/zh-CN.md
+++ b/blog/content/second-edition/posts/01-freestanding-rust-binary/zh-CN.md
@@ -6,16 +6,16 @@ date = 2018-02-10
 
 +++
 
-创建我们自己的操作系统内核的第一步是创建一个不链接标准库的Rust可执行文件。 这使得无需基础操作系统即可在[裸机]]上运行Rust代码。
+创建我们自己的操作系统内核的第一步是创建一个不链接标准库的Rust可执行文件。 这使得无需基础操作系统即可在[裸机][bare metal]上运行Rust代码。
 
-[裸机]: https://en.wikipedia.org/wiki/Bare_machine
+[bare metal]: https://en.wikipedia.org/wiki/Bare_machine
 
 <!-- more -->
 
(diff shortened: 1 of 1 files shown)


AntoineSebert:
I know you spent time working on these translations, but I think the maintainer should put his efforts in writing new parts of the OS and their dedicated posts instead of translating the existing ones.

And who is going to translate the rest of the posts to chinese, or to other languages ?
More important, if the posts are translated in other languages, how will the maintainer respond to their issue tickets, comments and pull requests if they are also written in those languages ?

It could be a good idea, but I would categorize it as a bit too ambitious and energy-consuming for this repository. "The Hell is paved of good intentions".

TheBegining:
You are right.

phil-opp:
@TheBegining Sorry for not replying earlier, I was very busy.

I already received some translation requests, mainly for Chinese and Japanese. Unfortunately, I don't know these languages, so it would be difficult for me to maintain the translated content (i.e. keeping it up to date and responding to issues etc.). However, maybe it's still a good idea to try given that there is so much interest. We could add a red warning on top of each translated post that says something like "This post is just a community translation and might be out of date or inaccurate. Please report any issues in English."

I try to investigate the new translation support of Zola (our static site generator) based on this PR in the `chinese-translation` branch.

Just for your information: There are already a few unofficial translations in other repositories, e.g. https://github.com/rustcc/writing-an-os-in-rust for Chinese.

Rustin-Liu:
If we are going to provide official translation support, I would love to contribute and support.

phil-opp:
@Rustin-Liu Awesome, thanks! I think we can try it as an experiment and see whether it works in practice.

phil-opp:
I opened https://github.com/phil-opp/blog_os/pull/692 to add experimental translation support to the blog. This pull request is already included in the `translations` branch, so we can close it.

PR #709, opened 2019-12-27
Reword sentence in first post

pamolloy:


blog/content/second-edition/posts/01-freestanding-rust-binary/index.md
```diff
@@ -145,7 +145,7 @@ Language items are special functions and types that are required internally by t
 [`Copy`]: https://doc.rust-lang.org/nightly/core/marker/trait.Copy.html
 [copy code]: https://github.com/rust-lang/rust/blob/485397e49a02a3b7ff77c17e4a3f16c653925cb3/src/libcore/marker.rs#L296-L299
 
-Providing own implementations of language items would be possible, but this should only be done as a last resort. The reason is that language items are highly unstable implementation details and not even type checked (so the compiler doesn't even check if a function has the right argument types). Fortunately, there is a more stable way to fix the above language item error.
+Providing our own implementations of language items would be possible, but this should only be done as a last resort. The reason is that language items are highly unstable implementation details and not even type checked (so the compiler doesn't even check if a function has the right argument types). Fortunately, there is a more stable way to fix the above language item error.
```
phil-opp: Thanks for the pull request! I'm not a native speaker, so I might be wrong, but adding "our" here doesn't feel ideal either.

How about rewording the sentence a bit:

```suggestion
While providing custom implementations of language items is possible, it should only be done as a last resort. The reason is that language items are highly unstable implementation details and not even type checked (so the compiler doesn't even check if a function has the right argument types). Fortunately, there is a more stable way to fix the above language item error.
```

What do you think?

phil-opp:
Thanks for the update and the clarifying comment!

PR #734, opened 2020-01-31
Make active section link more discreet

Menschenkindlein:
The feature is great!


phil-opp:
Thanks! I agree that it looks better this way.

PR #740, opened 2020-02-05
Recursive page tables require extensive TLB flushes in some cases

phil-opp:
Mentions an additional drawback in the discussion of recursive page tables.

Reported in https://github.com/phil-opp/blog_os/issues/570#issuecomment-556710283

cc @andyhhp


phil-opp:
This is a very old PR and I'm not sure if this is something that's worth mentioning. I think such large page table changes need a full TLB flush anyway.

PR #752, opened 2020-02-17
Add a language selector for browser-supported languages

phil-opp:
This displays a language selector if there are translations for the current page. The selector is only shown for content languages that the user enabled in their browser.

The selector is very rudimentary at the moment and only displays the language code. For the future, I would like to replace it with the proper name, e.g. Chinese (Simplified) instead of `zh-CN`.

### DIFF
--- a/blog/static/js/main.js
+++ b/blog/static/js/main.js
@@ -1,4 +1,6 @@
 window.onload = function() {
+  show_lang_selector();
+
   var container = document.querySelector('#toc-aside');
 
   if (container != null) {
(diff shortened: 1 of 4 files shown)


blog/static/js/main.js
```diff
@@ -61,3 +63,19 @@ function toc_scroll_position(container) {
...
   }
 }
+
+function show_lang_selector() {
+  var show_lang_selector = false;
+  for (language_selector of document.querySelectorAll('#language-selector li')) {
+    var lang = language_selector.getAttribute("data-lang-switch-to");
+    this.console.log(lang)
+    if (this.navigator.languages.includes(lang)) {
+      this.console.log("supported!");
```
bjorn3: Should these logs be removed?

phil-opp: Done in https://github.com/phil-opp/blog_os/commit/fa5b50075196326321444a04e9816e86e272132b. Thanks!

PR #756, opened 2020-02-23
Updated pc-keyboard to `0.5.0`

### LINKED ISSUE #742: Keyboard handling is slightly broken
If I rattle on the keyboard I get ControlLeft coming up regularly, even though I am far from the control keys. Upgrading pc_keyboard to 0.5.0 seems to fix it.

RKennedy9064:
I recently updated this in my project so I figured I'd help out since you do such great work. This should fix #742. The `keyboard::new` now takes 3 arguments, with the third being `pc_keyboard::HandleControl`. The two options are:
```
MapLettersToUnicode
If either Ctrl key is held down, convert the letters A through Z into Unicode chars U+0001 through U+001A. If the Ctrl keys are not held down, letters go through normally.

Ignore
Don't do anything special - send through the Ctrl key up/down events, and leave the letters as letters.
```
I left it `MapLettersToUnicode` for now since there's a match to handle Unicode characters, but it can be changed to `Ignore` if you prefer, just let me know.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,9 +73,9 @@ checksum = "72ef4a56884ca558e5ddb05a1d1e7e1bfd9a68d9ed024c21704cc98872dae1bb"
 
 [[package]]
 name = "pc-keyboard"
-version = "0.3.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fff50ab09ba31bcebc0669f4e64c0952fae1acdca9e6e0587e68e4e8443808ac"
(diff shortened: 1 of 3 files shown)


phil-opp:
Thanks for the PR!

> I left it `MapLettersToUnicode` for now since there's a match to handle Unicode characters, but it can be changed to `Ignore` if you prefer, just let me know.

I think I'd prefer `Ignore` because we currently don't handle the control characters, which the `MapLettersToUnicode` option produces. (We just print the characters to screen and the VGA text buffer does not know anything about the control characters.)

RKennedy9064:
Sounds good. I just updated it to use `Ignore` instead.

phil-opp:
Thanks!

PR #765, opened 2020-03-10
Removed unnecessary word

patrick-braun:
Removed the word "executes" from "The CPU executes tries to write to `0xdeadbeef`, which causes a page fault."


phil-opp:
Thanks!

PR #789, opened 2020-04-11
We no longer make entry points for Windows and macOS in post-01

ptrckd:
This line confused me. I thought I'd read past something but it looks like the content in post-01 that it is referring to changed in 0b9ca847356c3f932e038d3b32d5e85a6e53e971.

Thanks for the blog!


phil-opp:
Thank you!

PR #794, opened 2020-04-14
Fix command line in Chinese translation of post 2.

12101111:
https://github.com/phil-opp/blog_os/issues/792#issuecomment-613253056


phil-opp:
Thanks!

@Rustin-Liu While I generally agree with you, I don't really have a problem with that. Also, we can just squash-merge this PR to adjust the commit message.

phil-opp:
The CI failure is unrelated, therefore I'm force-merging this.

PR #804, opened 2020-05-18
Simplify executor by merging task_queue and wake_queue

phil-opp:
Proposed in https://github.com/phil-opp/blog_os/issues/768#issuecomment-629605586

This pull request also renames the `waiting_tasks` map to `tasks` and makes `run_ready_tasks` operate on `&mut Task` references instead of removing and re-inserting tasks from the map.

As another simplification, we now use the entry API on the `waker_cache` map instead of manual if conditions. Unfortunately this requires to destructure `self` into its fields because [RFC 2229]( https://github.com/rust-lang/rfcs/pull/2229) is not yet implemented (see the [tracking issue](https://github.com/rust-lang/rust/issues/53488)), but I still think that the new approach is simpler.

### DIFF
--- a/src/task/executor.rs
+++ b/src/task/executor.rs
@@ -4,24 +4,26 @@ use core::task::{Context, Poll, Waker};
 use crossbeam_queue::ArrayQueue;
 
 pub struct Executor {
+    tasks: BTreeMap<TaskId, Task>,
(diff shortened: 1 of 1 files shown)


phil-opp:
Post update in https://github.com/phil-opp/blog_os/pull/805.

PR #821, opened 2020-06-12
Update link to const evaluator documentation

rodrigovalle:
Thanks for the detailed blog posts, I'm really enjoying them. I noticed that this link has moved so I thought I'd update this post to point to the new location. Cheers!


phil-opp:
Thanks a lot!

PR #822, opened 2020-06-15
docs: remove version lock

hanhan9449:
fix error `error: the legacy LLVM-style asm! syntax is no longer supported`


hanhan9449:
sorry, I don't know how should I do.

phil-opp:
Thank you!

No worries about the "Zola Check" failure, that's just our external link checker that fails from time to time when some linked sites are not reachable.

PR #834, opened 2020-07-12
Fix grammar for freestanding-rust-binary linker arguments.

azjkjensen:
The grammar for MacOS linker arguments was clunky. Fix it to be grammatically correct.


phil-opp:
Thank you!

PR #835, opened 2020-07-16
Migrate post-02 code from cargo-xbuild to `-Zbuild-std`

phil-opp:
This pull request implements building using cargo's new [`build-std`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#build-std) instead of using the [`cargo-xbuild`](https://github.com/rust-osdev/cargo-xbuild) crate.

To enable the `build-std` functionality, we set the `unstable.build-std` configuration key in the `.cargo/config.toml` file. This flag was introduced very recently in https://github.com/rust-lang/cargo/pull/8393, so it only works on the latest nightlies. Unfortunately, rustfmt doesn't build on these nightlies yet, so you have to use `rustup update nightly --force` to install them (results in skipping the rustfmt component).

Functionality-wise, the `build-std` feature behaves almost identical to `cargo-xbuild`. The only difference is that it doesn't enable the `mem` feature of the `compiler_builtins` crate, which defines the `memcpy`, `memset`, etc operations. This is a [known issue](https://github.com/rust-lang/wg-cargo-std-aware/issues/53), which is hopefully fixed soon. Until then, we can work around it by adding a dependency on the [`rlibc`](https://docs.rs/rlibc/1.0.0/rlibc/) crate. To ensure that this crate is linked even though it is not directly used, we need to add an `extern crate rlibc` statement to our `main.rs`.

After these changes, our kernel is buildable through a normal `cargo build`. The `cargo run` and `cargo test` commands will also work as expected, provided the latest bootimage version is used.

This PR additionally rewrites our CI script based on the GitHub Actions provided by the [`actions-rs`](https://github.com/actions-rs/) organization. This makes Rust's error messages compatible with GitHubs annotations, so that any build errors and warnings are shown inline at the correct line.

The required changes to the blog post will be done in a separate follow-up PR.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,3 +1,5 @@
+unstable.build-std = ["core", "compiler_builtins"]
+
 [build]
 target = "x86_64-blog_os.json"
 
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,5 @@
-unstable.build-std = ["core", "compiler_builtins"]
+[unstable]
+build-std = ["core", "compiler_builtins"]
 
 [build]
 target = "x86_64-blog_os.json"
--- a/.github/workflows/build-code.yml
+++ b/.github/workflows/build-code.yml
@@ -49,11 +49,11 @@ jobs:
         with:
           profile: minimal
           toolchain: nightly
-      - name: "Install Rustup Components"
-        run: rustup component add rust-src llvm-tools-preview
       - name: "Install bootimage"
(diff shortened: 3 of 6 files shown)


phil-opp:
The additional changes required to the other `post-x` branches after merging this PR were minimal, so I pushed them directly.

phil-opp:
I submitted https://github.com/phil-opp/blog_os/pull/836 to update the blog posts for these changes.

PR #843, opened 2020-07-30
change rustup override add to rustup override set

RWOverdijk:
I'm not sure if this is a change you'd like, because it's technically not broken or anything. However, if you do, there are a couple of other files to change (4).

https://github.com/rust-lang/rustup/commit/33062a1a08eff4cbfd83027c09a69e496243738f

`add` is an alias and the cli help command also only lists `set` and `unset` for me (no add). For that reason I think it's better to use set (less confusing, in line with cli).


phil-opp:
Thank you! I agree that it makes sense to use the officially recommended (and documented) commands.

PR #849, opened 2020-08-18
Update index.md

pqnet:
To use the `write` macro you need to `use` the `core::fmt::Write` trait. Just importing `core::fmt` isn't enough.

### DIFF
--- a/blog/content/second-edition/posts/03-vga-text-buffer/index.md
+++ b/blog/content/second-edition/posts/03-vga-text-buffer/index.md
@@ -374,6 +374,7 @@ Now we can use Rust's built-in `write!`/`writeln!` formatting macros:
 
 ```rust
 // in src/vga_buffer.rs
+use core::fmt::Write;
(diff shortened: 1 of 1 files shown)


blog/content/second-edition/posts/03-vga-text-buffer/index.md
```diff
@@ -374,6 +374,7 @@ Now we can use Rust's built-in `write!`/`writeln!` formatting macros:
 
 ```rust
 // in src/vga_buffer.rs
+use core::fmt::Write;
 
 pub fn print_something() {
     use core::fmt::Write;
```
phil-opp, line 380: We import it here inside the function already.

phil-opp:
Closing as the import is already there (see https://github.com/phil-opp/blog_os/pull/849#discussion_r472782866). Thanks for opening this PR though! Let me know if you have problems with the import that is used in the post.

PR #850, opened 2020-08-19
Update Zola to 0.11.0

phil-opp:
This updates the blog to version 0.11.0 of Zola, which is the static site generator we use.

Most changes in Zola 0.11.0 are RSS feed related. This PR renames a few configuration keys to their new name and uses the `feed_filename` config key to keep our RSS file at `/rss.xml` to avoid breakage. Since the new version fixed anchor checking for uppercase HTML tags, we can remove some `zola check` exceptions.

### DIFF
--- a/.github/workflows/build-site.yml
+++ b/.github/workflows/build-site.yml
@@ -20,7 +20,7 @@ jobs:
     - uses: actions/checkout@v1
 
     - name: 'Download Zola'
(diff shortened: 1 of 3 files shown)


phil-opp:
Deployed in https://github.com/phil-opp/blog_os_deploy/commit/0c1cc5973ef809d45c710767fdb04bbf3b47c9a0. It's a large diff, but most changes come from different formatting of code blocks/spans and a change in the redirect template.

PR #854, opened 2020-09-09
Heap Allocation: Make wording of heap start prefix clearer

Hofer-Julian:


phil-opp:
Thanks for the PR! I think I meant that both addresses start with the `0x_4444_4444` prefix. Maybe something like `0x_4444_4444_*` would make it more clear?

Hofer-Julian:
Thanks for your comment @phil-opp. I understand what you meant to say now.
I've made a commit with your suggested changes.
But I think that simply closing this PR would be fine as well :)

phil-opp:
Thanks! I think the change makes things clearer, so let's merge it.

PR #856, opened 2020-09-13
add-move cursor base on print

zzhgithub:
add-move cursor base on print
see https://wiki.osdev.org/Text_Mode_Cursor
use llvm_asm


phil-opp:
Thanks and sorry for the late reply! I'm not sure if this is something that I want to add to the blog because this would require explaining inline assembly, I/O ports, and VGA registers very early in the series. While I agree that it's nice to be able to move the cursor, I'm not sure whether the additional complexity is worth it in such an early post.

phil-opp:
Closing since this won't be relevant for the upcoding 3rd edition (which uses a framebuffer instead of the VGA text mode).

Thanks for the PR!

PR #857, opened 2020-09-21
Add link to custom targets

briankung:
This makes it slightly easier to find out more about the JSON file that specifies custom targets.


phil-opp:
Thanks!

PR #864, opened 2020-10-01
Use shorthand initialization from earlier code

briankung:
Minor change, but keep the shorthand struct initialization seen previously at line 201: https://github.com/phil-opp/blog_os/blame/master/blog/content/second-edition/posts/03-vga-text-buffer/index.md#L201


phil-opp:
Thanks!

PR #870, opened 2020-10-08
Removed message about rustfmt

toothbrush7777777:
Removed message about recent Rust nightlies not containing `rustfmt`.


phil-opp:
Thanks!

Issue #872, opened 2020-10-21
add license owner and year


stmtk1:
apache license name and year is empty. So you should add your name.
<img width="519" alt="スクリーンショット 2020-10-21 10 39 46" src="https://user-images.githubusercontent.com/8833998/96662712-e9ddb300-1389-11eb-9941-609b640a86c5.png">

PR #873, opened 2020-10-21
Remove apache license appendix

stmtk1:
# issue 
- https://github.com/phil-opp/blog_os/issues/872
- apache license owner and year is empty. So you should add license onwer.
## before
スクリーンショット 2020-10-21 10 39 46
## after

スクリーンショット 2020-10-21 10 44 47


phil-opp:
It seems like the appendix isn't really required to be in the license file (see e.g. https://github.com/rust-lang/rust/pull/67734). So I would prefer to just remove the appendix.

stmtk1:
I removed apache lisence's appendix

phil-opp:
Thanks!

PR #875, opened 2020-10-30
Adding Right-to-Left support for template

hamidrezakp:
Adding Right-to-Left support for template. This is not a full solution to RTL problem because some English paragraphs are also Right-to-Left and must be fixed.

Now every post that want to be RTL must add an `rtl = true` in the `extra` section of post.

### DIFF
--- a/blog/templates/second-edition/page.html
+++ b/blog/templates/second-edition/page.html
@@ -33,11 +33,19 @@
 {% endblock toc_aside %}
 
 {% block main %}
+    {% if page.extra.rtl %}
+	<div class="right-to-left">
+    {% endif %}
+
     <h1>{{ page.title }}</h1>
     <time datetime="{{ page.date | date(format="%Y-%m-%d") }}" class="post-date">
         {{ page.date | date(format="%b %d, %Y") }}
         {% if page.extra.updated %} (updated on {{ page.extra.updated | date(format="%b %d, %Y") }}) {% endif %}
     </time>
+
+    {% if page.extra.rtl %}
+	</div>
+    {% endif %}
 
     {% if page.extra.warning %}
         <div class="warning">
@@ -67,7 +75,15 @@
         </div>
     {% endif %}
 
+    {% if page.extra.rtl %}
+	<div class="right-to-left">
+    {% endif %}
+
     {{ page.content | replace(from="<!-- toc -->", to=macros::toc(toc=page.toc))  | safe }}
+
+    {% if page.extra.rtl %}
+	</div>
+    {% endif %}
 
     <div class="post-footer-support">
         <h2>Support Me</h2>
--- a/blog/templates/second-edition/page.html
+++ b/blog/templates/second-edition/page.html
@@ -33,19 +33,13 @@
 {% endblock toc_aside %}
 
 {% block main %}
-    {% if page.extra.rtl %}
-	<div class="right-to-left">
-    {% endif %}
-
+    <div class="{% if page.extra.rtl %}right-to-left{% endif %}">
     <h1>{{ page.title }}</h1>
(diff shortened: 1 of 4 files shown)


phil-opp:
Thanks! Using an `extra.rtl` setting for specifying the text direction is a good idea.

Instead of adding a separate CSS file and applying the `rtl` layout to all text, I would prefer solving this via classes. We could add a `right-to-left` classes in our HTML and then use a rule like this in our existing `main.css`:

```css
.right-to-left p, .right-to-left h1, … {
	direction: rtl;
	font-family: "Vazir, Sahel, Yekan";
}
```

This should allow us to set the text direction per block. In the `page.html` template, we could then apply the text direction only to the `page.content` and `page.title`.

We could also add a `left-to-right` class that can be used for overriding the text direction within `right-to-left` blocks. Since CSS always applies the most specific rule, this should allow us to arbitrarily nest ltr and rtl blocks.

hamidrezakp:
Yeah you are right, using a `right-to-left` class in `main.css` is way better. also the `left-to-right` class is so good because we have some `code` blocks inside `p` elements and we can nicely use nested `rtl` and `ltr` in text.

However i don't know how we can apply `right-to-left` classes to `page.content`. i searched Zola documents but didn't find anything useful. can you give and example?
thanks

phil-opp:
We could just add an additional `<div>` element around it, to which we can add the class then.

hamidrezakp:
Ok, i have done changes based on `Classes`. let me know if there is any other issue.

blog/templates/second-edition/page.html
```diff
@@ -33,12 +33,20 @@ <h2>Table of Contents</h2>
...
     <h1>{{ page.title }}</h1>
     <time datetime="{{ page.date | date(format="%Y-%m-%d") }}" class="post-date">
         {{ page.date | date(format="%b %d, %Y") }}
         {% if page.extra.updated %} (updated on {{ page.extra.updated | date(format="%b %d, %Y") }}) {% endif %}
     </time>
 
+    {% if page.extra.rtl %}
+	</div>
+    {% endif %}
+
```
phil-opp, line 49: We can simplify this by making only the class of the div conditional:

```suggestion
    <div class="{% if page.extra.rtl %}right-to-left{% endif %}">
    <h1>{{ page.title }}</h1>
    <time datetime="{{ page.date | date(format="%Y-%m-%d") }}" class="post-date">
        {{ page.date | date(format="%b %d, %Y") }}
        {% if page.extra.updated %} (updated on {{ page.extra.updated | date(format="%b %d, %Y") }}) {% endif %}
    </time>
    </div>
```

blog/templates/second-edition/page.html
```diff
@@ -67,8 +75,16 @@ <h1>{{ page.title }}</h1>
...
+    {% if page.extra.rtl %}
+	<div class="right-to-left">
+    {% endif %}
+
     {{ page.content | replace(from="<!-- toc -->", to=macros::toc(toc=page.toc))  | safe }}
 
+    {% if page.extra.rtl %}
+	</div>
+    {% endif %}
+
```
phil-opp, line 87: Similar to above:

```suggestion
    <div class="{% if page.extra.rtl %}right-to-left{% endif %}">
    {{ page.content | replace(from="<!-- toc -->", to=macros::toc(toc=page.toc))  | safe }}
    </div>
```

phil-opp (commented):
Thanks for the update! I left some simplification suggestions inline:

hamidrezakp:
Thanks a lot for suggestions!, now code is cleaner and so straight forward.

phil-opp:
Thanks!

PR #877, opened 2020-11-01
Fix some links in 02-minimal-rust-kernel

Undin:
Previously:
"`mem` feature" led to wrong feature
"`memcpy` etc. implementations" had redundant parenthesis that prevented proper rendering


phil-opp:
Thanks!

PR #883, opened 2020-11-18
Emphasize moving code

briankung:
This prevents people from having duplicate code and potentially encountering errors. If both copies are kept the compiler may complain `error: cannot find macro `serial_println` in this scope` in `main.rs`

### DIFF
--- a/blog/content/second-edition/posts/04-testing/index.md
+++ b/blog/content/second-edition/posts/04-testing/index.md
@@ -687,7 +687,7 @@ To make the required functions available to our integration test, we need to spl
 
 Like the `main.rs`, the `lib.rs` is a special file that is automatically recognized by cargo. The library is a separate compilation unit, so we need to specify the `#![no_std]` attribute again.
 
-To make our library work with `cargo test`, we need to also add the test functions and attributes:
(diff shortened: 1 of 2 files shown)


blog/content/second-edition/posts/04-testing/index.md
```diff
@@ -687,7 +687,7 @@ To make the required functions available to our integration test, we need to spl
 
 Like the `main.rs`, the `lib.rs` is a special file that is automatically recognized by cargo. The library is a separate compilation unit, so we need to specify the `#![no_std]` attribute again.
 
-To make our library work with `cargo test`, we need to also add the test functions and attributes:
+To make our library work with `cargo test`, we need to also move the test functions and attributes to `lib.rs` from `main.rs`:
```
phil-opp, line 690: Minor suggestion:

```suggestion
To make our library work with `cargo test`, we need to also move the test functions and attributes from `main.rs`  to `lib.rs`:
```

phil-opp (commented):
Thanks! I left one small suggestion inline, otherwise this looks good to me!

phil-opp:
Thanks!

PR #888, opened 2020-12-10
Fix broken QEMU link

JohnTitor:
Use specified version instead (it'd be great if Zola could ignore some paths on link-checker, I imagine).


phil-opp:
Thanks!

> it'd be great if Zola could ignore some paths on link-checker

It can and we're already using that feature: https://github.com/phil-opp/blog_os/blob/e8f5ee95c6867cb782418bd727e8f7c6f3e6d25b/blog/config.toml#L19-L30

In general I try to check as many links as possible in order to prevent link-rot.

PR #890, opened 2020-12-16
Rename edition folders

phil-opp:
To prepare for the planned third edition of the blog.

No existing links should change. Most pages use a custom path anyway and for the few that don't I added aliases.

In addition to renaming the content folders, this PR also duplicates the CSS and Javascript files for each edition. This way, we are able to completely change the page design for the third edition if we like, e.g. to add a dark mode.


phil-opp:
The diff at https://github.com/phil-opp/blog_os_deploy/commit/cf05aba576da32ee6cc6404163c9fa59afebbfc7 looks as expected and confirms that no HTML pages changed their path without redirect.

PR #894, opened 2020-12-31
Minor grammar fixes

delta1:
2 small fixes 

"we don't have implemented" -> "we haven't implemented"  

and 

"We use lazy_static again, because Rust's const evaluator is not powerful enough yet" -> "As before, we use lazy_static again." 

just to remove what has already been said a few paragraphs up. 

Thank you for this amazing resource!


hamidrezakp:
Nice, its better now

phil-opp:
Thank you!

PR #896, opened 2020-12-31
fix layout of right-to-left in homepage and tables

hamidrezakp:
The layout of Home page had problem showing right-to-left texts so i fixed it.
Also tables in posts that are right-to-left had the same issue causing show them incorrectly.


phil-opp:
Thank you!

PR #897, opened 2021-01-02
Improve explanation for `BootInfoFrameAllocator::usable_frames`

delta1:
Minor rewrite of the `flap_map` explanation of frame allocator, removing the reference to `into_iter` that is no longer in the code. 

Hope that helps!


phil-opp:
Thank you!

PR #898, opened 2021-01-02
remove unused import

delta1:
Removes the unused imports of serial_print/ln from heap allocation example code


phil-opp:
Thanks!

PR #910, opened 2021-01-14
Added note to specify version 0.2.6 of volatile

NickSchmitt:
Added explicit note to specify `volatile` crate version `0.2.6`, since later versions produce an error.

### DIFF
--- a/blog/content/edition-2/posts/03-vga-text-buffer/index.md
+++ b/blog/content/edition-2/posts/03-vga-text-buffer/index.md
@@ -305,8 +305,7 @@ We can add a dependency on the `volatile` crate by adding it to the `dependencie
 volatile = "0.2.6"
 ```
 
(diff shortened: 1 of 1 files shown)


bjorn3:
I think the dependency should be updated instead.

NickSchmitt:
[phil-opp says he's rewriting the section.](Referencing discussion here https://github.com/phil-opp/blog_os/issues/405#issuecomment-703905698)

phil-opp (commented):
Thanks! I left one small phrasing suggestion inline, otherwise this looks good to me.

Yes, I think that sticking to version 0.2 is the best solution for now. The newer version 0.3 uses a completely different interface so updating to it would require some effort. I rather invest that time into the upcoming third edition.

NickSchmitt:
Thank you that makes a lot of sense.

phil-opp:
Thanks!

PR #912, opened 2021-01-27
Fix typo

EvanMu96:
Hi. There was a trivial JSON tailing comma problem in edition-2, post-02, zh-CN(only). So I fixed that for consistency across languages.


phil-opp:
Thanks!

PR #913, opened 2021-01-27
Translate post-08 to Persian

MHBahrampour:


phil-opp:
Thanks for the pull request! Is this ready to be merged?

MHBahrampour:
I think so, we checked it once but maybe there is some minor issues that we didn't catch.
But i think it is ready to be merged.
-------- Original Message --------

phil-opp:
Ok great, thanks a lot!

PR #932, opened 2021-02-23
Remove note on builtin memory optimizations

dspencer12:
The referenced issue in compiler-builtins (https://github.com/rust-lang/compiler-builtins/pull/365) has been merged. Hence, I think it's safe to remove the note about the memory implementations not being optimized.

I'm not sure if you would want to state that the implementations are optimized - I think it's likely assumed by the reader unless otherwise stated.


phil-opp:
Thanks!

PR #943, opened 2021-03-10
post-06: Add spaces around some "two asterisk" notations

woodyZootopia:
@garasubo CC: @JohnTitor 
In some cases, the markdown rendering engine of `zola` fails in rendering **bold** (two-asterisk) notations.
For example, it should be rendered like this:
2021-03-10T14:36:23,652703242+09:00
but is rendered like this right now:
2021-03-10T14:27:20,969354582+09:00

This seems to be happening because Japanese doesn't use spaces around words. In order to fix this, I add spaces around the notation in those cases. I'm not sure this is the best solution, but this at least makes the rendering engine parse the notation correctly.


woodyZootopia:
I also found a minor mistake in translation of TSS（タスクステートセグメント）, so fixed it in the second commit.

garasubo:
Thanks for finding out those problem. LGTM.

phil-opp:
Thanks!

PR #952, opened 2021-03-24
Example image does not match code output

alexxroche:
The example code prints `L4 ` in the l4_table loop but the example image does not.


phil-opp:
Thanks!

PR #953, opened 2021-03-24
x86_64::VirtAddr already imported on L415

alexxroche:
`error[E0252]: the name 'VirtAddr' is defined multiple times`


phil-opp:
Thanks!

PR #959, opened 2021-03-30
Simplify the later example error

alexxroche:
Required so that the current compiler output match the example error on line 637, (because the compiler is now clever enough to warn about):
```bash
warning: variable does not need to be mutable
 --> src/main.rs:5:9
  |
5 |     let mut heap_value = Box::pin(SelfReferential {
  |         ----^^^^^^^^^^
  |         |
  |         help: remove this `mut`
  |
  = note: `#[warn(unused_mut)]` on by default
```
Which is distracting. (Also a chance to passively introduce `#[allow(unused_mut)]` )  

Though this proposed change may, in itself be a distraction for the student?

### DIFF
--- a/blog/content/edition-2/posts/12-async-await/index.md
+++ b/blog/content/edition-2/posts/12-async-await/index.md
@@ -549,6 +549,7 @@ Using heap allocation, we can try to create a self-referential struct:
 
 ```rust
 fn main() {
+    #[allow(unused_mut)]
     let mut heap_value = Box::new(SelfReferential {
         self_ptr: 0 as *const _,
     });


phil-opp:
Thanks for the PR! I'm not sure if allowing the lint is the best solution. I think it would make more sense to either add the "does not need to be mutable" warning to the output or to make the `heap_value` binding non-mut.

alexxroche:
I think your judgement is best. I did think about the non-mut option, but didn't know if changing that would be a distraction for the student. (Just as long as the presented output matches the actual output, I'm sure it will be fine how you chose to fix this.)

PR #963, opened 2021-04-02
wake_trait is now stable

alexxroche:
You probably know: `wake_trait` has been stable since 1.51.0 and no longer requires an attribute to enable


phil-opp:
Thanks!

phil-opp:
(Note to self: `post-12` branch updated in a9e8caed904bb1d0b153512e357e0fd93a076c44)

PR #964, opened 2021-04-02
Previously updated dependency

alexxroche:
Previously, in https://os.phil-opp.com/testing/ we set the dependency to `x86_64 = "0.13.2"` so this advice seems redundant.


phil-opp:
Good catch, thanks!

PR #974, opened 2021-04-22
Suggestion for .cargo/config.toml Explanation

moomoolive:
I was confused on what the 'unstable' section of my .cargo/config.toml file should look like once I started getting linking errors when going through the third post. It turned out it was because I overwrote the previous 'unstable' configuration with the latter one. 

In other words, I thought we were to overwrite "build-std = ["core", "compiler_builtins"]" under the "unstable" config, with "build-std-features = ["compiler-builtins-mem"]" and NOT have both present. 

I think this makes it more clear that both are supposed to be present.


phil-opp:
Thanks, I agree that this makes things more clear!

PR #983, opened 2021-05-11
Remove wrong suggestion part

HKalbasi:
Issue https://github.com/rust-lang/rust/issues/68049 is now fixed!

So we can remove that part. It isn't on nightly currently but will come. [Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=1eca2c196142cd9060a157d1f296d8d1) for checking it.


HKalbasi:
It is correct on nightly now.

phil-opp:
Great, thanks a lot!

HKalbasi:
image
What should I do? :)

phil-opp:
This was probably caused by yesterday's [Github outage](https://www.githubstatus.com/incidents/zbpwygxwb3gw). Sorry for the confusion!

PR #1031, opened 2021-06-30
docs: update version of crates

### LINKED ISSUE #1030: Apply #990 to the other languages
Update the version of following crates.

- `x86_64`
- `pic8259` (from `pic8259_simple`)

It seems `linked_list_allocator` is not included in the translated posts yet.

#990

Foo-x:
Closes #1030

Applied #990 to the other languages.


phil-opp:
Thanks a lot! I think that the update requires some more changes, e.g. changing `&mut InterruptStackFrame` to `InterruptStackFrame`. I think this mainly affects the Japanese and Persian translations (cc @hamidrezakp @MHBahrampour @woodyZootopia @JohnTitor).

Foo-x:
Thanks to your review!
I checked it, and it didn't work without the changes as you said.
Fixed.

hamidrezakp:
> Thanks a lot! I think that the update requires some more changes, e.g. changing `&mut InterruptStackFrame` to `InterruptStackFrame`. I think this mainly affects the Japanese and Persian translations (cc @hamidrezakp @MHBahrampour @woodyZootopia @JohnTitor).

I think there is no problem with Persian translation after this commit.
Did you mean the problem with Right-to-left?

JohnTitor (approved):
I think it'd be great if we could also update `translation_based_on_commit` but otherwise this applies the same changes as #990, LGTM.

phil-opp:
Thanks everyone!

PR #1036, opened 2021-07-14
docs: fix typo

Foo-x:
Fixed typos in post-12.


phil-opp:
Thanks! I don't speak Japanese, but Google translate shows only a minimal difference for this, so I'm gonna trust you on that.

PR #1072, opened 2022-01-24
Add additional “real hardware” instructions

kennystrawnmusic:
Because macOS and Linux are both Unix-like operating systems, they both have the `dd` command bundled. During my 13 years of having Linux as my daily driver desktop OS and 1 year of having a MacBook Pro on the side, I’ve used this command long enough to know that it definitely gets this job done. So, time to start contributing.


phil-opp:
Looks good, thank you! One small suggestion: Could you use a placeholder name that is less likely to exist? For example `sdXX` instead of `sdb`. People often copy some commands without reading the surrounding text, so we the command to fail if copied verbatim.

kennystrawnmusic:
Done. The Mac version also needed similar treatment, so changed both.

phil-opp:
Thanks!

PR #1073, opened 2022-01-25
Add screenshot of `wrapping_add` output

kennystrawnmusic:
Doing even more to help contribute to the composition of this


phil-opp:
Thanks! Instead of linking to an external image, could you commit the image in-tree? Just copy it to the same folder as the `index.md`. Also, a more descriptive alt text would be nice.

kennystrawnmusic:
Done

phil-opp:
Thanks! I think we could still improve the alt text a bit. The current "QEMU showing an increasing value on each loop iteration" is not really a description of the image, it rather describes the code. Maybe something like "shows multiple columns, where each column is a gradient from black to white" or something like that.

phil-opp:
Thanks!

PR #1082, opened 2022-02-23
Adding fix to make the test_runner functions pub

TornaxO7:
I always got the following error message when I tried chapter 4:

```rust
error[E0463]: can't find crate for `test`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `os` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

After comparing your source code on branch `post_04` with my code, I found out that the issue was function signature of `test_runner`: My signature was without `pub`. Adding the `pub` fixed it for me.

**HINT:** I don't really know if this PR is "necessary" because I couldn't find anyone else in the comments who got the same problem. Could it be that it just happened on my machine?

Greetings from Kerlsruhe by the way :)

### DIFF
--- a/blog/content/edition-2/posts/04-testing/index.fa.md
+++ b/blog/content/edition-2/posts/04-testing/index.fa.md
@@ -77,7 +77,7 @@ error[E0463]: can't find crate for `test`
 #![test_runner(crate::test_runner)]
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


phil-opp:
Sorry, this somehow slipped through my notifications. Thanks for submitting!

PR #1083, opened 2022-02-26
Fix wrong wording

jdmichaud:
Was saying:
"This is the approach that we will use for building our OS kernel"
if my understanding is correct, should say:
"This is **not** the approach that we will use for building our OS kernel"


phil-opp:
I agree that the wording is a bit confusing there. What I meant to say is that we will use a bare-metal target instead of passing certain linker arguments (the next section). So we _do_ use that approach, with a slight variation (a custom target).

phil-opp:
Closing because I think that the current wording is correct. I'm happy to merge a PR to make this less confusing though.

PR #1089, opened 2022-03-06
Fix french in 01-freestanding-rust-binary

alaincao:
Hi, just spotted those 2 mistakes in french while following the blog in french


phil-opp:
Thanks! I don't speak French, so maybe @AlexandreMarcq or @dallenng can take a quick look?

AlexandreMarcq:
@phil-opp those are indeed mistakes. My bad! Thanks @alaincao !

phil-opp:
Thanks for checking!

PR #1107, opened 2022-05-17
Change Unpin to Pin

asami-kawasaki:
## summary
Isn't the part "Unpin" in 【】 the following sentence "Pin"?
Sorry if I am wrong.

## Changes
Before change
```
By wrapping `Self` into `Pin` and letting the compiler opt-out of 【Unpin】 for self-referential futures generated from async/await, it is guaranteed that the futures are not moved in memory between `poll` calls. This ensures that all internal references are still valid.
```

After change
```
By wrapping `Self` into `Pin` and letting the compiler opt-out of `Pin` for self-referential futures generated from async/await, it is guaranteed that the futures are not moved in memory between `poll` calls. This ensures that all internal references are still valid.
```

### DIFF
--- a/blog/content/edition-2/posts/12-async-await/index.ja.md
+++ b/blog/content/edition-2/posts/12-async-await/index.ja.md
@@ -701,7 +701,7 @@ unsafe {
 fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output>
 ```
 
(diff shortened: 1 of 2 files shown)


bjorn3:
No, `Unpin` is a trait implemented for most types by default. It allows you to wrap and unwrap `Pin` from values of such types.

PR #977, opened 2021-05-02
Translate post-12 to Japanese

kahirokunn:
The title says it all.
@JohnTitor Could you review the translation?

PR #1108, opened 2022-05-17
Correction of Japanese

asami-kawasaki:
## summary
The following Japanese has been corrected because it has a different meaning.

## corresponding passage
Before change
```
コンパイラに `Unpin` を【選択】させることで、
```

After change
```
コンパイラに `Unpin` を【オプトアウト】させることで、
```

### DIFF
--- a/blog/content/edition-2/posts/12-async-await/index.ja.md
+++ b/blog/content/edition-2/posts/12-async-await/index.ja.md
@@ -701,7 +701,7 @@ unsafe {
 fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output>
 ```
 
-このメソッドが通常の`&mut self`ではなく`self: Pin<&mut Self>`を取る理由は、[上][self-ref-async-await]で見たように、async/awaitから生成されるfutureのインスタンスはしばしば自己参照しているためです。`Self` を `Pin` にラップして、async/await から生成された自己参照のfutureに対して、コンパイラに `Unpin` を選択させることで、`poll` 呼び出しの間にfutureがメモリ内で移動しないことが保証されます。これにより、すべての内部参照が有効であることが保証されます。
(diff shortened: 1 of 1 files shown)


kahirokunn:
LGTM

phil-opp:
Thank you! And thanks to @kahirokunn for reviewing!

One thing that I'm not sure about is the translators list. The current list only includes the people that substantially contributed to the initial translation added in https://github.com/phil-opp/blog_os/pull/977, but not people who contributed minor corrections afterwards (e.g. in https://github.com/phil-opp/blog_os/pull/1036). These contributions are still very valuable of course, the question is just if we want to add them to the same list as the initial translators.

So we have two possibilities, which both have their advantages and drawbacks:

- Add all contributors to the list, independent of their contibutions
  - All contributions are credited
  - The list will then grow very long over time
  - The initial translators are equally credited as people who fixed a typo
- Only add the initial translators to the list and leave it unchanged on later corrections.
  - Maybe we could have a separate list for more minor contributions, e.g. named `contributors`
  - We need to decide whether a contribution is "major enough" to be credited

What do you think?

cc'ing the other people in the list: @kahirokunn @garasubo @sozysozbot @woodyZootopia

kahirokunn:
Contributions like this one are also important and will only be a better resource if they are actively made.

Maybe contributions like this one are best written on a separate list.

kahirokunn:
I hope many people can correct mistranslations like this one.

I think solved by the design of the website is better.

kahirokunn:
I explain how this occurred.
I asked `asami-kawasaki` to review the translation of this article.
`asami-kawasaki` had read the translation in detail, found several corrections, and contributed to it.
So, I said to `asami-kawasaki` put `asami-kawasaki` in the list of translators.
asami-kawasaki just did the job as instructed.

garasubo:
I don't have strong opinion on it. Either way is fine for me.

phil-opp:
Ok, thanks a lot for your thoughts! Sounds like a separate contributors list is the best solution. I'll implement this in the template, then I will merge this PR.

> Contributions like this one are also important and will only be a better resource if they are actively made.

Yes, absolutely! Contributions like this are very valuable and we definitely want to encourage them. I hope that my earlier comment did not come across otherwise... So thanks a again for your work @asami-kawasaki and sorry for this whole discussion!

phil-opp:
I added the new field in #1112 and merged this PR as part of #1113.

PR #1113, opened 2022-05-22
Merge correction for Japanese translation of post-12 and use `translation_contributors` field

phil-opp:
Merge #1108 and use new translation contributors field introduced in https://github.com/phil-opp/blog_os/pull/1112.

@Foo-x I also added you to the new contributors list because you contributed to this translation in the past. I hope this is ok with you.


Foo-x:
Of course it's ok! Thank you for adding me.

PR #1116, opened 2022-06-07
Move comment threads for translated posts to separate category

phil-opp:
Avoids that a search for the post title returns a thread for a translated post.


phil-opp:
Follow-up fix in eeac9e880c70d4b8fd76ed9101fc801cc6073723

PR #1123, opened 2022-07-07
Check writing of all posts in edition 2

hecatia-elegua:
This fixes grammar, punctuation, etc. and clarifies some sentences.

I'm keeping this as a draft for now, but I want to help out a lot more. If you want, I can also split every writing check into its own PR, but I'm trying to stay conservative with the changes.
I'm using my common sense and, in some places, grammar checkers.

### DIFF
--- a/blog/content/edition-2/posts/05-cpu-exceptions/index.md
+++ b/blog/content/edition-2/posts/05-cpu-exceptions/index.md
@@ -180,9 +180,10 @@ On a normal function call (using the `call` instruction), the CPU pushes the ret
 
 For exception and interrupt handlers, however, pushing a return address would not suffice, since interrupt handlers often run in a different context (stack pointer, CPU flags, etc.). Instead, the CPU performs the following steps when an interrupt occurs:
 
+0. **Saving the old stack pointer**: The CPU reads the stack pointer (`rsp`) and stack segment (`ss`) register values and remembers them in an internal buffer.
 1. **Aligning the stack pointer**: An interrupt can occur at any instruction, so the stack pointer can have any value, too. However, some CPU instructions (e.g., some SSE instructions) require that the stack pointer be aligned on a 16-byte boundary, so the CPU performs such an alignment right after the interrupt.
 2. **Switching stacks** (in some cases): A stack switch occurs when the CPU privilege level changes, for example, when a CPU exception occurs in a user-mode program. It is also possible to configure stack switches for specific interrupts using the so-called _Interrupt Stack Table_ (described in the next post).
-3. **Pushing the old stack pointer**: When the interrupt occurs (before the alignment), the CPU pushes the values of the stack pointer (`rsp`) and stack segment (`ss`) registers. This makes it possible to restore the original stack pointer when returning from an interrupt handler.
+3. **Pushing the old stack pointer**: The CPU pushes the `rsp` and `ss` values from step 0 to the stack. This makes it possible to restore the original stack pointer when returning from an interrupt handler.
 4. **Pushing and updating the `RFLAGS` register**: The [`RFLAGS`] register contains various control and status bits. On interrupt entry, the CPU changes some bits and pushes the old value.
 5. **Pushing the instruction pointer**: Before jumping to the interrupt handler function, the CPU pushes the instruction pointer (`rip`) and the code segment (`cs`). This is comparable to the return address push of a normal function call.
 6. **Pushing an error code** (for some exceptions): For some specific exceptions, such as page faults, the CPU pushes an error code, which describes the cause of the exception.
--- a/blog/content/edition-2/posts/05-cpu-exceptions/index.md
+++ b/blog/content/edition-2/posts/05-cpu-exceptions/index.md
@@ -8,7 +8,7 @@ date  = 2018-06-17
 chapter = "Interrupts"
 +++
 
-CPU exceptions occur in various erroneous situations, for example when accessing an invalid memory address or when dividing by zero. To react to them we have to set up an _interrupt descriptor table_ that provides handler functions. At the end of this post, our kernel will be able to catch [breakpoint exceptions] and to resume normal execution afterwards.
+CPU exceptions occur in various erroneous situations, for example, when accessing an invalid memory address or when dividing by zero. To react to them, we have to set up an _interrupt descriptor table_ that provides handler functions. At the end of this post, our kernel will be able to catch [breakpoint exceptions] and resume normal execution afterward.
 
 [breakpoint exceptions]: https://wiki.osdev.org/Exceptions#Breakpoint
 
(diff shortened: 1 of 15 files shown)


hecatia-elegua:
This is ready now :)

02, 10, 11, and 12 have some more info in their commit messages about where I changed some outdated stuff.

phil-opp:
Thanks a lot for all your work! I'm still reviewing your changes, but what I've seen so far looks very good.

blog/content/edition-2/posts/05-cpu-exceptions/index.md
```diff
@@ -180,12 +180,12 @@ On a normal function call (using the `call` instruction), the CPU pushes the ret
...
+1. **Aligning the stack pointer**: An interrupt can occur at any instruction, so the stack pointer can have any value, too. However, some CPU instructions (e.g., some SSE instructions) require that the stack pointer be aligned on a 16-byte boundary, so the CPU performs such an alignment right after the interrupt.
+2. **Switching stacks** (in some cases): A stack switch occurs when the CPU privilege level changes, for example, when a CPU exception occurs in a user-mode program. It is also possible to configure stack switches for specific interrupts using the so-called _Interrupt Stack Table_ (described in the next post).
+3. **Pushing the old stack pointer**: When the interrupt occurs (before the alignment), the CPU pushes the values of the stack pointer (`rsp`) and stack segment (`ss`) registers. This makes it possible to restore the original stack pointer when returning from an interrupt handler.
```
phil-opp, line 185: To me, this seems to change the meaning. The "at the time when the interrupt occurred (before the alignment)" of the original wording refers to the values of the `rsp` and `ss` registers. So the CPU does not push their current value (which might be changed in steps 1 and 2), but it pushes the old values that these registers had at the time of the interrupt.

Your new wording sounds like the push happens before the alignment. However, the push happens after the alignment and the stack switching, it just pushes the original, unchanged register values.

blog/content/edition-2/posts/06-double-faults/index.md
```diff
@@ -476,15 +476,15 @@ fn stack_overflow() {
...
 
 [volatile]: https://en.wikipedia.org/wiki/Volatile_(computer_programming)
 [`Volatile`]: https://docs.rs/volatile/0.2.6/volatile/struct.Volatile.html
 [_tail call elimination_]: https://en.wikipedia.org/wiki/Tail_call
 
-In our case, however, we want that the stack overflow happens, so we add a dummy volatile read statement at the end of the function, which the compiler is not allowed to remove. Thus, the function is no longer _tail recursive_ and the transformation into a loop is prevented. We also add the `allow(unconditional_recursion)` attribute to silence the compiler warning that the function recurses endlessly.
+In our case, however, we want the stack overflow to happen, so we add a dummy volatile read statement at the end of the function, which the compiler cannot remove. Thus, the function is no longer _tail recursive_, and the transformation into a loop is prevented. We also add the `allow(unconditional_recursion)` attribute to silence the compiler warning that the function recurses endlessly.
```
phil-opp, line 487: > which the compiler _cannot_ remove.

I'm not sure if I like the _cannot_ here. It sounds a bit like the compiler is not capable enough to remove it. However, the only reason that the compiler keeps the statement is that the languages rules explicitly forbid removing volatile instructions.

blog/content/edition-2/posts/12-async-await/index.md
```diff
@@ -844,11 +844,11 @@ impl Task {
 }
 ```
 
-Since the [`poll`] method of the `Future` trait expects to be called on a `Pin<&mut T>` type, we use the [`Pin::as_mut`] method to convert the `self.future` field of type `Pin<Box<T>>` first. Then we call `poll` on the converted `self.future` field and return the result. Since the `Task::poll` method should be only called by the executor that we create in a moment, we keep the function private to the `task` module.
+Since the [`poll`] method of the `Future` trait expects to be called on a `Pin<&mut T>` type, we use the [`Pin::as_mut`] method to convert the `self.future` field of type `Pin<Box<T>>` first. Then we call `poll` on the converted `self.future` field and return the result. Since the `Task::poll` method should only be called by the executor that we create in a moment, we keep the function private to the `task` module.
```
luisgizirian, line 847: Shouldn't be a `we'll` instead of `we` in [..."method should be only called by the executor that **we** create in a moment"...]

luisgizirian: Rephrasing the original paragraph fix proposal to:`be called only (or: exclusively) by`

blog/content/edition-2/posts/11-allocator-designs/index.md
```diff
@@ -1230,7 +1230,7 @@ There are many more allocator designs with different tradeoffs. [Slab allocation
 
 ## What's next?
 
-With this post, we conclude our memory management implementation for now. Next, we will start exploring [_multitasking_], starting with [_threads_]. In subsequent post we will then explore [_multiprocessing_], [_processes_], and cooperative multitasking in the form of [_async/await_].
+With this post, we conclude our memory management implementation for now. Next, we will start exploring [_multitasking_], starting with cooperative multitasking in the form of [_async/await_]. In subsequent posts, we will then explore [_threads_],[_multiprocessing_], and [_processes_].
```
phil-opp, line 1233: Good catch!

Minor spacing issue:

```suggestion
With this post, we conclude our memory management implementation for now. Next, we will start exploring [_multitasking_], starting with cooperative multitasking in the form of [_async/await_]. In subsequent posts, we will then explore [_threads_], [_multiprocessing_], and [_processes_].
```

blog/content/edition-2/posts/12-async-await/index.md
```diff
@@ -169,11 +169,11 @@ let file_content = loop {
...
+A more efficient approach could be to _block_ the current thread until the future becomes available. This is, of course, only possible if you have threads, so this solution does not work for our kernel, at least not yet. Even on systems where blocking is supported, it is often not desired because it turns an asynchronous task into a synchronous task again, thereby inhibiting the potential performance benefits of parallel tasks.
 
 #### Future Combinators
 
-An alternative to waiting is to use future combinators. Future combinators are methods like `map` that allow chaining and combining futures together, similar to the methods on [`Iterator`]. Instead of waiting on the future, these combinators return a future themselves, which applies the mapping operation on `poll`.
+An alternative to waiting is to use future combinators. Future combinators are methods like `map` that allow chaining and combining futures together, similar to the methods in [`Iterator`]. Instead of waiting on the future, these combinators return a future themselves, which applies the mapping operation on `poll`.
```
phil-opp, line 176: "in" sounds a bit strange to me here, maybe "the methods _of_ `Iterator`" or "the methods of the `Iterator` trait"?

phil-opp (commented):
Awesome work, thanks a lot for taking the time to go through all the posts!

I left a few comments, but overall this looks all very good to me.

blog/content/edition-2/posts/08-paging-introduction/index.md
```diff
@@ -49,18 +49,18 @@ By modifying the memory addresses before the actual access, segmentation already
 
 ### Virtual Memory
 
-The idea behind virtual memory is to abstract away the memory addresses from the underlying physical storage device. Instead of directly accessing the storage device, a translation step is performed first. For segmentation, the translation step is to add the offset address of the active segment. Imagine a program accessing memory address `0x1234000` in a segment with offset `0x1111000`: The address that is really accessed is `0x2345000`.
+The idea behind virtual memory is to abstract away the memory addresses from the underlying physical storage device. Instead of directly accessing the storage device, a translation step is performed first. For segmentation, the translation step is to add the offset address of the active segment. Imagine a program accessing memory address `0x1234000` in a segment with an offset `0x1111000`: The address that is really accessed is `0x2345000`.
```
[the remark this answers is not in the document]
hecatia-elegua: Oops, yes, mine was wrong, forgot the "of". Your original works too, but the alternative sounds a bit more.. fluent?

blog/content/edition-2/posts/05-cpu-exceptions/index.md
hecatia-elegua: Funny enough, I wrote something about this down as a note while I was reading:

> "before the alignment" -> should that be step 0 instead?

So it sounded to me like "the CPU pushes `rsp` and `ss` before the alignment" (which would make more sense as step 0) or "the CPU pushes `rsp` and `ss` but with the value from before the alignment" (where you could leave out the last part, the next sentence makes it clear)

Now reading your comment, I still don't seem to get it. Sorry. Could you suggest a new sentence for step 3, written from scratch? Might just be some sort of ambiguity I don't get.

phil-opp: Maybe it makes sense to split this one into two steps:

- step 0: Read the `rsp` and `ss` values and remember in an internal buffer.
- steps 1 and 2 as before
- step 3: Push the `rsp` and `ss` values from step 0 to the stack.

So the push happens in step 3, but the values that are written are from step 0.

hecatia-elegua:
Thank you for this amazing project!

phil-opp (approved):
Thanks! :)

PR #1130, opened 2022-07-29
Grammar fix

luisgizirian:
Post 12. `we'll` instead of `we` in [..."method should be only called by the executor that **we'll** create in a moment"...]

This is a very educational project. Tough subject, written in an easy to follow and digestible way. Thanks for putting it out @phil-opp!

### DIFF
--- a/blog/content/edition-2/posts/12-async-await/index.md
+++ b/blog/content/edition-2/posts/12-async-await/index.md
@@ -844,7 +844,7 @@ impl Task {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


luisgizirian:
I've checked #1123 and looks like this has slipped over that major corrections proposal.

luisgizirian:
Closing this one as I moved this small fix proposal into #1123.

phil-opp:
Thank you for helping to improve this series! 

> This is a very educational project. Tough subject, written in an easy to follow and digestible way. Thanks for putting it out

Thanks a lot, great to hear that :).

Issue #1145, opened 2022-10-03
Post-09 Paging Implementation: Address Calculation section has a bad URL for image


seewishnew:
The first image in address calculation section of paging implementation appears to have a bad URL:
![image](https://user-images.githubusercontent.com/14972559/193522246-bb0393b8-4bf8-43d6-8680-ac33f1509460.png)

Changing the URL to `../paging-introduction/x86_64-table-indices-from-address.svg` (removing the `08-` prefix) fixes the issue:
![image](https://user-images.githubusercontent.com/14972559/193522514-c44ce68d-91e1-42c2-8e0e-a9460bde19b6.png)

phil-opp:
Fixed in https://github.com/phil-opp/blog_os/pull/1146

PR #1146, opened 2022-10-03
Fixes bad URL from post-09 address calculation section

seewishnew:
Fixes bad image URL in post-09 paging implementation/address-calculation section as detailed in: https://github.com/phil-opp/blog_os/issues/1145


phil-opp (approved):
Good catch, thank you!

PR #1151, opened 2022-10-03
Remove warning output from QEMU command.

lovemeforareason:
The command to run QEMU produces an output, but the output and
its message is not needed to complete the tutorial.


phil-opp:
Thanks! I agree that we don't need to include the warning there.

Issue #1138, opened 2022-09-09
panicked at 'assertion failed: max_phys_addr < (1 << 48) / 512' in post-09


willothy:
Crashed when running my code, and confirmed occurring as well with the code from `post-09` branch.

It does not appear to be using my own panic handler (won't print text other than the error), even though other panics do. 

Running Arch Linux, Rustc 1.65 nightly, QEMU 7.1.0. Same error occurs on Windows.

6th gen i7, 48GB RAM.

Error does not appear on another PC which is running a 10th gen i7 (QEMU 7.0.0).

abusch:
Seeing the same thing on Macos with `rustc 1.65.0-nightly (9243168fa 2022-08-31)`. It seems to happen as soon as I enable the `map_physical_memory` feature of the `bootloader` crate.

phil-opp:
Thanks for reporting!

We currently see a test failure on our macOS CI runners as well. When I compared the logs for the last-working run and the first failing one, I noticed that QEMU updated from v7.0.0 to v7.1.0. The runners for Ubuntu and Windows stil use much older QEMU versions and don't show any problems.

Since Arch Linux is also typically running the newest software, I could imagine that this is triggered by the QEMU v7.1.0 release. Could you check your QEMU versions to see if my suspicion is correct?

phil-opp:
The assertion failure happens here: https://github.com/rust-osdev/bootloader/blob/42da77b47558478c5a948d27cac80767b484eacc/src/main.rs#L289-L291 . If the assertion fails, it means that the memory map contains a memory region at address >512GiB, which is a bit strange. I think you should be able to bypass the assertion by creating a `[package.metadata.bootloader]` table in your `Cargo.toml` and add a `physical-memory-offset = "0x0000f00000000000"` field to it (choose any address you like). Then you should be able to inspect the memory map in your kernel. Maybe this tells us the reason for this assertion failure.

abusch:
I can confirm I have QEmu 7.1.0 installed.

I used your workaround which indeed removed the panic, and was able to print the memory map:

```
Memory map: [
    MemoryRegion {
        range: FrameRange(0x0..0x1000),
        region_type: FrameZero,
    },
    MemoryRegion {
        range: FrameRange(0x1000..0x5000),
        region_type: PageTable,
    },
    MemoryRegion {
        range: FrameRange(0x5000..0x17000),
        region_type: Bootloader,
    },
    MemoryRegion {
[...]

PR #1176, opened 2022-11-25
Fixes bootloader version

seewishnew:
This PR upgrades the bootloader crate's version from 0.9.8 to 0.9.23 to account for this fix: https://github.com/rust-osdev/bootloader/pull/264, which can affect post-09 examples. 

There is also a note about using the bootloader's custom physical memory offset feature (from https://github.com/phil-opp/blog_os/issues/1138#issuecomment-1242711282) and some of its caveats.


phil-opp:
Thanks!

PR #1207, opened 2023-04-07
Ammend small typo: Change from set_cs to set_reg

dporr:
`set_cs` is deprecated according to: https://docs.rs/x86_64/latest/x86_64/instructions/segmentation/fn.set_cs.html

Also in the post the displayed code uses `x86_64::instructions::segmentation::CS::set_reg` instead of `x86_64::instructions::segmentation::set_cs`

### DIFF
--- a/blog/content/edition-2/posts/06-double-faults/index.md
+++ b/blog/content/edition-2/posts/06-double-faults/index.md
@@ -374,7 +374,7 @@ pub fn init() {
 }
 ```
 
-We reload the code segment register using [`set_cs`] and load the TSS using [`load_tss`]. The functions are marked as `unsafe`, so we need an `unsafe` block to invoke them. The reason is that it might be possible to break memory safety by loading invalid selectors.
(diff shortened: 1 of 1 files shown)


blog/content/edition-2/posts/06-double-faults/index.md
```diff
@@ -374,7 +374,7 @@ pub fn init() {
 }
 ```
 
-We reload the code segment register using [`set_cs`] and load the TSS using [`load_tss`]. The functions are marked as `unsafe`, so we need an `unsafe` block to invoke them. The reason is that it might be possible to break memory safety by loading invalid selectors.
+We reload the code segment register using [`set_reg`] and load the TSS using [`load_tss`]. The functions are marked as `unsafe`, so we need an `unsafe` block to invoke them. The reason is that it might be possible to break memory safety by loading invalid selectors.
 
 [`set_cs`]: https://docs.rs/x86_64/0.14.2/x86_64/instructions/segmentation/fn.set_cs.html
```
bjorn3, line 379: This link needs to be updated.

phil-opp:
Thanks a lot!

The change was merged as part #1218, which also included the link update.

PR #1216, opened 2023-04-29
The `#[alloc_error_handler]` attribute was removed

phil-opp:
Allocation errors now always panic.

See:
- https://github.com/rust-lang/rust/pull/102318
- https://github.com/rust-lang/rust/pull/109507


phil-opp:
Merged into `post-11` and `post-12` as well.

PR #1218, opened 2023-05-04
posts/06-double-faults/index: fixup deprecated fn

twilfredo:
The code segment from the example uses the newer function whilst the text segment refers to the old/deprecated `set_cs` fn. Let's fix that.

Note: that the method reference link points to docs 0.14.5 and not 0.14.2. It looks like it doesn't exist in the older version (?).

### DIFF
--- a/blog/content/edition-2/posts/06-double-faults/index.md
+++ b/blog/content/edition-2/posts/06-double-faults/index.md
@@ -374,9 +374,9 @@ pub fn init() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


twilfredo:
I just realized there's a pending PR for it! But it's missing the updated link which this patch contains, see https://github.com/phil-opp/blog_os/pull/1207

phil-opp (approved):
Thanks!

Issue #1215, opened 2023-04-26
Misaligned Pointer panic in the Paging Introduction post


PitchBlackNights:
I don't know if this is very important, but it's been driving me crazy trying to fix it.

I'm almost at the end of the [Introduction to Paging](https://os.phil-opp.com/paging-introduction). I'm at the test to see if reading and writing to the code segment address will cause a paging fault.

When I run the binary, it spits out:
```
panicked at 'misaligned pointer dereference: address must be a multiple of 0x4 but is 0x204f63', src\main.rs:18:22
```
Here is the code in question (I'm guessing that it's really complaining about line 16, which is the first line included in this snippet):
```rust
let ptr = 0x204f63 as *mut u32;

unsafe { let x = *ptr; }
println!("Read Worked");

unsafe { *ptr = 42; }
println!("Write Worked");
```

I've gotten this error on previous uses of `0xdeadbeaf` as a virtual address, so I would always just swap those out for unnecessarily large addresses like `0x4444444444`.

I did look to see if anyone else was having this error, and I did find [one person](https://github.com/phil-opp/blog_os/discussions/1011#discussioncomment-5623435).

TeeJaey:
I would like to comment that I am also running into this issue. 

I so far always swapped `0xdeadbeaf` for `0xdeadbea0` since that is a multiple of 0x4.

While that works, I am worried I might run into bigger problems later down the line.

carloalbertogiordano:
Hi, maybe the read_unaligned and the write_unaligned could be used in this case. I have not encountered this error so far.
Here are some links:
https://doc.rust-lang.org/std/ptr/fn.read_unaligned.html
https://doc.rust-lang.org/std/ptr/fn.write_unaligned.html

PitchBlackNights:
Okay, thanks @bjorn3 ! I haven't tried it, but I think it will work. I'll leave this open for now though, as it should be working without read/write_unaligned

bjorn3:
Dereferencing unaligned pointers is UB. Previously if you were lucky the compiler wouldn't miscompile it, but to catch this UB, rustc recently added a compiler pass when debug assertions are enabled which adds code to check on every pointer dereference if the pointer is aligned or not and if not aborts. This helps finding this UB.

PitchBlackNights:
Okay, so that's just how rustc compiles it now. Thanks!

PR #1226, opened 2023-05-20
Fix panics caused by misaligned pointer dereferences in "Double Faults" & "Introduction to Paging"

SPuntte:
**EDIT 2023-05-31:**

I just noticed that post 8 is affected by this as well (as already discussed in #1215) and applied the same fix there.

___

In the "Double Faults" post, the following code is given as an example:
```Rust
// trigger a page fault
unsafe {
    *(0xdeadbeef as *mut u64) = 42;
};
```

Dereferencing `0xdeadbeef` as any type larger than `u8` is [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html), and [a recently merged PR for Rust](https://github.com/rust-lang/rust/pull/98112) added runtime debug assertions for this.

Thus, instead of the desired effects, the above code panics with the message `'misaligned pointer dereference: address must be a multiple of 0x8 but is 0xdeadbeef'`.

Fortunately, the author chose the value `42` such that it fits in an `u8`, facilitating an easy fix without the need to invent another funny address that would additionally be 8-byte aligned ;)

The English blog text does not mention the type of the pointer so I assume that neither do the translations.

### DIFF
--- a/blog/content/edition-2/posts/06-double-faults/index.fa.md
+++ b/blog/content/edition-2/posts/06-double-faults/index.fa.md
@@ -49,7 +49,7 @@ pub extern "C" fn _start() -> ! {
 
     // trigger a page fault
     unsafe {
-        *(0xdeadbeef as *mut u64) = 42;
+        *(0xdeadbeef as *mut u8) = 42;
     };
 
     // as before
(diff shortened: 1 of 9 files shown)


riking:
> without the need to invent another funny address that would additionally be 8-byte aligned ;)

abad1dea is 4-byte aligned.

PR #1234, opened 2023-08-16
Update post 2 with beginner friendly cargo tips

Connortsui20:
As a relatively new person to Rust, I confused the `.cargo/config.toml` with the global cargo config in my home directory (`~/.cargo/config.toml). While this is pretty obviously wrong in hindsight, since I've never used the `[unstable]` options before, I didn't realize my mistake until this next thing that tripped me up.

For `cargo install bootimage`, I think it is reasonable to tell the reader to go into a different directory to execute the command, since it might be the case that the reader has never dealt with different targets before and would have no idea that running `cargo install` for the new target that they just made in a json would be wrong (at least this was me).

This could be worded differently than the changes I made, but I think that the addition of these could only benefit a confused reader.

I'm new to PRs in general so apologies if I messed up something with the branches!

### DIFF
--- a/blog/content/edition-2/posts/02-minimal-rust-kernel/index.md
+++ b/blog/content/edition-2/posts/02-minimal-rust-kernel/index.md
@@ -260,10 +260,10 @@ That's where the [`build-std` feature] of cargo comes in. It allows to recompile
 [`build-std` feature]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#build-std
 [nightly Rust compilers]: #installing-rust-nightly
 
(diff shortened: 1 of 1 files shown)


phil-opp (commented):
Thanks a lot for the PR, these are some useful clarifications! The only thing that I'm not sure about is the `blog_os` prefix in the paths. We don't really use this convention anywhere else, so I fear that this might lead to some confusion too. For example, I could imagine that readers create a blog_os subdirectory within their project folder (which might have a different name).

Connortsui20:
Yeah I definitely agree, maybe the above change would make sense? I don't know if it is too wordy though.

phil-opp (approved):
Thanks!

PR #1262, opened 2024-01-05
fix testing and paging introduction chapter errors in zh-CN translation

acyanbird:
I don't know if this fit the format hhh


phil-opp (approved):
Thanks a lot!

PR #1269, opened 2024-01-27
Fix typos in code for `embedded_graphics` crate in chapter 3

proudmuslim-dev:


blog/content/edition-3/posts/03-screen-output/index.md
```diff
@@ -312,22 +312,24 @@ Fortunately, there is the nice `no_std`-compatible [`embedded-graphics`] crate,
...
 
-    fn draw_pixel(&mut self, pixel: Pixel) {
+    fn draw_pixel(&mut self, Pixel(coordinates, color): Pixel<Rgb888>) {
         // ignore any pixels that are out of bounds.
         let (width, height) = {
             let info = self.framebuffer.info();
             (info.width, info.height)
         }
+
         if let Ok((x @ 0..width, y @ 0..height)) = coordinates.try_into() {
```
proudmuslim-dev, line 333: This still won't compile on account of the fact that the `Point` type apparently doesn't implement `Into<(usize, usize)>`. Attempting to change the type just results in more issues because the range values must be of the same integer type and have to be constants. The checks could of course be moved down but that would look much worse

proudmuslim-dev: I should also note that this syntax requires you to enable the experimental `exclusive_range_pattern` feature in order for it to be valid in the first place

phil-opp (approved):

blog/content/edition-3/posts/03-screen-output/index.md
phil-opp: Thanks! This is already a good improvement, so let's merge this. I try my best to continue working on finishing the post soon.

PR #1272, opened 2024-01-28
Update edition3/03/index.md

xuanplus:
enable it to print "Hello Rust!" on qemu:

屏幕截图(1)

I find that I can't use runtime values inside the pattern, I can only match on a range made of compile-time constants.

Therefore, I add the function `in_range` to judge it.

The `Point` from embedded_graphics can't use `try_into`? So I add a `From` to change it to Position.

The `impl OriginDimensions` is from example of embedded_graphics official:

> This trait should be implemented instead of [Dimensions](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) if the top left corner of the bounding box will always be at the origin, which will be the case for most display drivers. Some types, like [ImageDrawable], require a bounding box that starts at the origin and can only be used if [OriginDimensions](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) is implemented.

Anyway, now, It works.

### DIFF
--- a/blog/content/edition-3/posts/03-screen-output/index.md
+++ b/blog/content/edition-3/posts/03-screen-output/index.md
@@ -157,6 +157,15 @@ pub struct Position {
     pub y: usize,
 }
 
+impl From<Point> for Position {
+    fn from(value: Point) -> Self {
+        Self {
+            x: value.x as usize,
+            y: value.y as usize,
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks for submitting this! I will probably not merge this directly, but I will definitely consider it when I write the rest of this post.

PR #1274, opened 2024-01-29
change some format errors for chinese edition minimal kernel

acyanbird:
Some hyperlinks are not displaying properly.


phil-opp (approved):
Thanks a lot!

PR #1275, opened 2024-01-29
typo

lachsdachs:
"doubly" doesn't seem like a real word


proudmuslim-dev:
It is in fact a real word- I'd argue doubly is more appropriate here

lachsdachs:
> It is in fact a real word- I'd argue doubly is more appropriate here

oh god you are right, i should have looked that up

PR #1312, opened 2024-04-13
Fix some minor error in zh-tw

squirrel532:
Hello, here is what i found to improve the zh-tw translation of ch1. 

1. Fix some typo in zh-tw.
2. Change translate of specific terms.
  - name mangling -> 名稱重整
  - calling convention -> 呼叫慣例
3. Remove needless words.


phil-opp:
Thanks!

PR #1340, opened 2024-08-12
04-testing: Added panic-abort-tests flag note to "duplicate lang item" workaround

Fincap:
`panic-abort-tests = true` will, as described by the [cargo unstable features docs](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#panic-abort-tests), enable nightly support to compile test harness crates with `-Cpanic=abort`. Without this flag Cargo will compile tests, and everything they depend on, with `-Cpanic=unwind` because it’s the only way test-the-crate knows how to operate.

This fixed my compilation issues that were not solved by the other suggested workaround.


phil-opp (approved):
Thanks a lot!

PR #1349, opened 2024-10-06
Update post 02 "Building our Kernel" section

EsotericDryad:
Added/corrected the explanation for why the new target still uses Linux conventions


phil-opp (approved):
Thank you!

PR #1369, opened 2025-01-01
Correcting many_boxes_long_lived test that incorrectly passes for BumpAllocator when run with cargo test -r

Robert-M-Lucas:
`BumpAllocator` currently passes `many_boxes_long_lived` in release mode due to `long_lived` and the `assert_eq` being optimised away. Testing in release mode can be useful as it can make tests run faster overall, despite the longer compile time per test and ensures correct functionality when debug code isn't included.

### DIFF
--- a/tests/heap_allocation.rs
+++ b/tests/heap_allocation.rs
@@ -7,6 +7,7 @@
 extern crate alloc;
 
 use alloc::{boxed::Box, vec::Vec};
+use core::hint::black_box;
 use blog_os::allocator::HEAP_SIZE;
 use bootloader::{entry_point, BootInfo};
 use core::panic::PanicInfo;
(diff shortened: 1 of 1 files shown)


AtomicGamer9523:
Is `BumpAllocator ` supposed to fail `many_boxes_long_lived` in debug mode ?
If so, then shouldn't the test be annotated as such ?
And if not, then is there any evidence that `long_lived` gets optimized away ?

Robert-M-Lucas:
@AtomicGamer9523  The `BumpAllocator` is meant to fail `many_boxes_long_lived` (as it does in debug mode when you are going through the tutorial) as `long_lived` prevents any freeing which it does not do in release mode with `cargo test --release`. While it is not the final allocator used, the `BumpAllocator` remains in the source code and its passing of this test shows that the test does not correctly test other allocators in release mode, something that I found useful to run tests faster (particularly when increasing the heap size).

AtomicGamer9523:
Interesting, my main concern with `black_box` is [this](https://doc.rust-lang.org/core/hint/fn.black_box.html#:~:text=Programs%20cannot%20rely,for%20correctness%2C). There must be a better way to tell the compiler to not optimize out `long_lived`.

Wisma-55 (approved):

ViktorPopp (approved):

PR #1377, opened 2025-01-15
Use new `&raw const` operator instead of `unsafe {& _}`

phil-opp:


phil-opp:
Code branches updated in aa9d0fff.

Issue #906, opened 2021-01-12
Tracking issue for Japanese translations


JohnTitor:
This is a tracking issue for Japanese translations. If you have any questions/problems with Japanese translations, feel free to leave a comment here!

phil-opp:
I just updated the 'Heap Allocation' post to remove the `alloc_error_handler`, which was removed from `rustc`:

- https://github.com/phil-opp/blog_os/pull/1217

It would be great if someone could apply the same change to the Japanese translation of the post. Thanks!

JohnTitor:
I'll take a look next week unless someone beats me.

phil-opp:
Note: I prepared https://github.com/phil-opp/blog_os/pull/1500 to update the translations with the latest changes from the English originals again. I don't speak the languages, so I used Claude Code for this. It would be great if someone could take a quick look at my changes to the Japanese translations. Thanks!

PR #1386, opened 2025-01-30
Fix Japanese translation of "run in parallel"

yo-goto:
ref #906

This post seems to have a translation error.  
I've corrected it.


phil-opp (approved):
Thank you!

PR #1399, opened 2025-02-20
Fix typo when referencing the `rustc-abi` field.

felixpackard:
Post 02 currently states...

> As we want to use the `soft-float` feature, we also need to tell the Rust compiler `rustc` that we want to use the corresponding ABI. We can do that by setting the `x86-softfloat` field to `x86-softfloat`.

...when the field name should be `rustc-abi`.

I've updated this for all localizations.


phil-opp (approved):
Thank you!

PR #1407, opened 2025-04-02
Update index.zh-CN.md

JINHUILYU:
Fix translation inconsistency: Update incorrect term "组建" to the correct term "组件" at line 645.


phil-opp (approved):
Thanks!

PR #1412, opened 2025-05-01
Added [[bin]] section to Cargo.toml

tigeryant:
...with test and bench set to false to prevent duplicate definition of lang item


phil-opp (approved):
Thank you!

PR #1420, opened 2025-06-25
translate edition2@post-11 to Chinese

ttttyy:


phil-opp:
Thanks a lot for your work! 

It seems like no one who speaks Chinese has time for a review. So let's merge this as-is, sorry for the delay!

PR #1422, opened 2025-07-31
Fix broken link in 04-testing

imaspacecat:
the wikipedia link in "lots of UART models" wasn't working anymore, so I changed it to the newer link


imaspacecat:
I didn't even change those links :sob:

PR #1436, opened 2025-09-03
Fix: `target-pointer-width` field now expects an integer

phil-opp:
- Update to latest bootloader version
- Fix: `target-pointer-width` field now expects an integer

See https://github.com/rust-lang/rust/pull/144443


phil-opp:
Merged into other post branches

PR #1462, opened 2026-02-16
Post-3 traduction in French

YaogoGerard:
Traduction of the part "vga-text-mode" in french


phil-opp (approved):
Thanks a lot!

I pushed a commit to list you as the author of the translation, I hope this is ok with you. Please let me know if not, then I'll remove it again.

YaogoGerard:
Thanks , That suits me.

PR #1463, opened 2026-02-17
add test flag in post 1&4 cn&jp

ic3w1ne:
Notice that new parts about test flag in Cargo.toml in post 1 and 4. Add their Chinese and Japnese translations.


phil-opp (approved):
Thank you!

PR #1477, opened 2026-05-03
fix: lazy_static is not needed, keyboard methods are const, 

harsh-98:
compile time initialization is possible for keyboard.

pub const fn Keyboard::new
pub const fn ScancodeSet1::new


phil-opp (approved):
Good catch, thanks!

phil-opp:
Applied to subsequent post branches

handregister:
Unsubscribe me please.

phil-opp:
Blog update in #1482

PR #1479, opened 2026-05-09
use spin::Lazy instead of lazy_static

NikoPit:
because lazy_static is depreceated


sebthirion:
`Lazy` changed name to `LazyLock` in the [spin 0.11.0 release](https://codeberg.org/zesterer/spin/src/commit/5197822543aa66f30a6cae4cb94a5ddfe5d4b1cb/CHANGELOG.md?trans-rights=human-rights#changed-2). Unless there's a good reason to stay on 0.10, we might as well bump to the latest release and use the new name.

phil-opp:
Thanks for opening this and sorry for the late reply!

I didn't find any official deprecation announcement from `lazy_static`. I only found https://github.com/rust-lang-nursery/lazy-static.rs/issues/214, which is a proposal to deprecate it, but it looks like no decision has been made at this point. 

I still agree that it would be a good idea to replace this crate because it's unmaintained and it's always good to replace special macro syntax with standard Rust code IMO.

PR #1484, opened 2026-07-06
Fix target spec: use `softfloat` instead of `x86-softfloat`

phil-opp:
The compatibility alias `x86-softfloat` was removed in https://github.com/rust-lang/rust/pull/157151.


phil-opp:
Merged into subsequent post branches

phil-opp:
blog update in https://github.com/phil-opp/blog_os/pull/1485

PR #1493, opened 2026-07-20
Update `x86_64` dependency to v0.15

phil-opp:
`x86_64` v0.14 does no longer work with recent Rust nightlies because the `Step` trait was changed.


phil-opp:

PR #1494, opened 2026-07-20
Update GDT and VirtAddr code to x86_64 v0.15

phil-opp:
- The `add_entry` method is named `append` now
- `VirtAddr` no longer impls `Add<usize>`

Follow-up to https://github.com/phil-opp/blog_os/pull/1493


phil-opp:

PR #1497, opened 2026-07-20
Fix: `VirtAddr` does no longer implement `Add<usize>`

phil-opp:
Part of the upgrade to x86_64 v0.15 (https://github.com/phil-opp/blog_os/pull/1493)


phil-opp:

